PR #3, opened 2024-06-11
Add requests needed for web into the account.

kluver:
This PR adds the functions currently used by poprox web into this repository.

### DIFF
--- a/src/poprox_storage/repositories/accounts.py
+++ b/src/poprox_storage/repositories/accounts.py
@@ -127,7 +127,7 @@ class DbAccountRepository(DatabaseRepository):
         subscription_tbl = self.tables["subscriptions"]
 
         create_query = subscription_tbl.insert().values(account_id=account_id)
-        if self.fetch_subscription(account_id) is None:
+        if self.fetch_subscription_for_account(account_id) is None:
             self.conn.execute(create_query)
 
     def end_subscription_for_account(self, account_id):
--- a/src/poprox_storage/repositories/accounts.py
+++ b/src/poprox_storage/repositories/accounts.py
@@ -53,6 +53,16 @@ class DbAccountRepository(DatabaseRepository):
             return accounts[0]
         return None
 
+    def create_new_account(self, email: str) -> Account:
+        account_tbl = self.tables["accounts"]
+        query = (
+            sqlalchemy.insert(account_tbl)
+            .values(email=email)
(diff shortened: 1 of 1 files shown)


src/poprox_storage/repositories/accounts.py
```diff
@@ -53,6 +53,16 @@ def fetch_account_by_email(self, email: str) -> Optional[Account]:
...
 
+    def create_new_account(self, email: str) -> Account:
+        account_tbl = self.tables["accounts"]
+        query = (
+            sqlalchemy.insert(account_tbl)
+            .values(email=email)
+            .returning(account_tbl.c.account_id, account_tbl.c.email)
+        )
+        row = self.conn.execute(query).one_or_none()
+        return Account(account_id=row.account_id, email=row.email)
```
kluver, line 64: work for a new PR, but I can confirm `row.variable` works. The documentation also says that `row["variable"]` works. This should save us from having to use `row[index]` which is more fragile.

src/poprox_storage/repositories/accounts.py
```diff
@@ -104,3 +114,30 @@ def fetch_subscribed_accounts(self):
         )
         account_ids = self._id_query(account_query)
         return self.fetch_accounts(account_ids)
+
+    def fetch_subscription_for_account(self, account_id):
```
karlhigley, line 118: I think have type annotations is going to help us keep track of which methods do what in this early realm where we don't have consistent method names yet, so I'd probably add some on these methods

karlhigley (approved):

src/poprox_storage/repositories/accounts.py
kluver: Well. -- I was a bit unsure on this one. In theory account_id is a UUID, but I see it as str elsewhere in this file, and that doesn't seperate the different types of UUIDs. As for return value -- it's technically just a raw row which is currently only used as None/not None. 

I'm guessing if we really wanted to fix the return value that would be doable, but might involve actually changing the return value (should be doable without problematic side-effect) for the parameters, we should probably pick a "canonical" representation of an id.

karlhigley: The canonical representation should be `UUID` with conversion to `str` as needed

kluver: Hmm... I think fixing this will have to be another PR -- I found multiple `str` in there and decided to standardize against what was already there.

PR #7, opened 2024-06-17
Kluver/feature/account onboarding

kluver:
Depends on: CCRI-POPROX/poprox-platform#76 and CCRI-POPROX/poprox-concepts#5

- Updates methods for creating account to store an account source.
- Adds methods for updating account status
- Adds methods for recording consent

### DIFF
--- a/src/poprox_storage/repositories/accounts.py
+++ b/src/poprox_storage/repositories/accounts.py
@@ -36,7 +36,7 @@ class DbAccountRepository(DatabaseRepository):
         account_tbl = self.tables["accounts"]
 
         query = select(
-            account_tbl.c.account_id, account_tbl.c.email, account_tbl.source
+            account_tbl.c.account_id, account_tbl.c.email, account_tbl.c.status
         )
         if account_ids is not None:
             query = query.where(account_tbl.c.account_id.in_(account_ids))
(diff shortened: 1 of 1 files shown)


src/poprox_storage/repositories/accounts.py
```diff
@@ -145,3 +156,19 @@ def end_subscription_for_account(self, account_id: UUID):
...
         self.conn.execute(delete_query)
+
+    def record_consent(self, account_id: UUID, document_name: str):
+        consent_tbl = self.tables["account_consent_log"]
+        query = sqlalchemy.insert(consent_tbl).values(
+            account_id=account_id, document_name=document_name
+        )
+        self.conn.execute(query)
+
+    def update_status(self, account_id: UUID, new_status: str):
```
karlhigley, line 167: Would it make sense to do these two database operations as part of the same transaction?

karlhigley (approved):

src/poprox_storage/repositories/accounts.py
kluver: I think it would. I don't really know how transactions are best managed in this data design. If it's just a connection property you can easily make a transaction before building the repository object and then commit after right?

PR #8, opened 2024-06-19
entity name lookup matches on lowercase

ShahanM:
There were crashes due to case mismatch between the form entry and db entry.


karlhigley (approved):

kluver (commented):
I'm a bit suspect of this -- I prefer true string match where possible. But I also can't see any immediate harm here, so I think we're probably good.

PR #10, opened 2024-07-01
Refactor `Article` and `Experiment` repositories to clean up their inserts

karlhigley:
Aside from style changes to appease the linter, this does a few things:
* Simplifies the process of inserting Pydantic models into the database in `Repository` classes
* Cleans up uses of the `upsert_and_return_id` function, which should be calls to the `_upsert_and_return_id` method
* Moves single-use code that inserts experiment stuff into the database from standalone public functions to private methods

### DIFF
--- a/src/poprox_storage/repositories/data_stores/db.py
+++ b/src/poprox_storage/repositories/data_stores/db.py
@@ -32,6 +32,17 @@ class DatabaseRepository:
         result = self.conn.execute(query).fetchall()
         return [row[0] for row in result]
 
+    def _insert_model(self, table_name: str, model, *, constraint_name: str | None = None, exclude=None):
+        if not constraint_name:
+            constraint_name = "uq_" + table_name
+
+        return self._upsert_and_return_id(
+            self.conn,
+            self.tables[table_name],
+            model.model_dump(exclude=exclude),
+            constraint=constraint_name,
+        )
+
     def _upsert_and_return_id(
         self,
         conn,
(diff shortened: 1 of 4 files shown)


src/poprox_storage/repositories/data_stores/db.py
```diff
@@ -32,48 +32,54 @@ def _id_query(self, query):
...
+        addl_fields: dict[str, Any] | None = None,
+        *,
+        constraint: str | None = None,
+        exclude=None,
+        commit: bool = True,
+    ):
+        fields: dict[str, Any] = model.model_dump(exclude=exclude)
+
+        if addl_fields:
+            fields.update(addl_fields)
```
karlhigley, line 48: For any fields that have different names from the database columns, we can use `exclude` to omit them from the insert and `addl_fields` to add them back in with a different name

src/poprox_storage/repositories/experiments.py
```diff
@@ -148,6 +122,77 @@ def get_active_expt_allocations(self, date: datetime.date | None = None) -> dict
...
+            {
+                "phase_id": phase_id,
+                "group_id": treatment.group.group_id,
+                "recommender_id": treatment.recommender.recommender_id,
+            },
+            exclude={"group", "recommender"},
+            commit=False,
+        )
+
+    def _insert_expt_assignment(
```
karlhigley, line 183: This method follows the pattern of the methods above that each take a Pydantic model and a foreign key id, but I cheated here and passed a `Group` object that's only used to grab the group id. This code was written before the `Allocation` object had been created, but it should probably be used here.

kluver (approved):

PR #14, opened 2024-07-12
Move database migrations and dev database setup from `poprox-platform`

karlhigley:
This is a full copy of the contents of the `poprox-db` and `dev` directories, with only minor modifications:
* The dev dependencies have been moved from `requirements-dev.txt` into `pyproject.toml`
* The README has been updated to contain instructions for installing the dev dependencies

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,10 +8,16 @@
 **Table of Contents**
 
 - [Installation](#installation)
-- [License](#license)
+
 
 ## Installation
 
 ```console
 pip install poprox-storage
+```
+
+### Dev Dependencies
+
+```console
+pip install .[dev]
 ```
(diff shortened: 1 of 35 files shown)


README.md
```diff
@@ -8,10 +8,16 @@
...
 ## Installation
 
 ```console
 pip install poprox-storage
 ```
+
+### Dev Dependencies
+
+```console
+pip install .[dev]
```
kluver, line 22: wait. This works? I always do `pip install -r <filename>`

dev/dump.sql
kluver, line 1: this is listed as an empty file. I don't think the original version of this was empty. If it was that's a problem on my end.

README.md
kluver: Ahh -- this works with the changes in pyproject right?

kluver (approved):
One question which I think is more TIL for kluver than actual uncertainty about correctness.

Another "this file shouldn't be empty" situation which I think is ultimately _fine_ if it is empty -- we can just do a new DB dump if we need to -- I don't think that needs to block this PR.

dev/dump.sql
karlhigley: The version in `poprox-platform` was empty too—I thought that was an intentional placeholder for a dump to be added manually

PR #15, opened 2024-07-15
Expand article fields fetched by repo to include recent additions

karlhigley:


kluver (approved):
lgtm

kluver:
I just double-checked that the table column names all match -- and they should. Double LGTM

PR #17, opened 2024-07-15
Add get click within timeframe function and write unit test

sophiasun0515:
This change is dependent on the new `Click` object created in https://github.com/CCRI-POPROX/poprox-concepts/pull/12/.

We aim to add the feature to get clicks for a list of users within certain timeframe.

New unit test passed locally. Also added an `example-test.py` file in the `dev` folder for future database query development reference.

### DIFF
--- a/src/poprox_storage/repositories/clicks.py
+++ b/src/poprox_storage/repositories/clicks.py
@@ -35,7 +35,7 @@ class DbClicksRepository(DatabaseRepository):
         super().__init__(connection)
         self.tables = self._load_tables("clicks")
 
-    def track_click_in_database(self, newsletter_id, account_id, article_id, created_at):
+    def track_click_in_database(self, newsletter_id, account_id, article_id, created_at=None):
         click_table = self.tables["clicks"]
         with self.conn.begin():
             stmt = insert(click_table).values(
--- a/src/poprox_storage/repositories/clicks.py
+++ b/src/poprox_storage/repositories/clicks.py
@@ -4,7 +4,8 @@ from collections import defaultdict
 from uuid import UUID
 
 from poprox_concepts import Account, ClickHistory
-from sqlalchemy import insert, select
+from poprox_concepts.domain.click import Click
+from sqlalchemy import insert, select, and_
 
 from poprox_storage.aws import s3
 from poprox_storage.aws.exceptions import PoproxAwsUtilitiesException
@@ -64,5 +65,32 @@ class DbClicksRepository(DatabaseRepository):
         histories = {}
         for account_id, user_clicks in clicked_articles.items():
             histories[account_id] = ClickHistory(article_ids=user_clicks)
+
+        return histories
+
+    def get_clicks_within_timestamp(self, accounts: list[Account], start_time, end_time) -> dict[UUID, list[Click]]:
+        click_table = self.tables["clicks"]
+
+        click_query = select(click_table.c.account_id, click_table.c.article_id, click_table.c.created_at).where(
+            and_(
+                click_table.c.account_id.in_([acct.account_id for acct in accounts]),
+                click_table.c.created_at >= start_time,
+                click_table.c.created_at <= end_time
+            )
+        )
+        click_result = self.conn.execute(click_query).fetchall()
+
+        clicked_articles = defaultdict(list)
+        for row in click_result:
+            clicked_articles[row[0]].append(Click(article_id=row[1], timestamp=row[2]))
+        
+        for account in accounts:
+            account_id = account.account_id
+            if account_id not in clicked_articles:
+                clicked_articles[account_id] = []
+        
+        histories = {}
+        for account_id, user_clicks in clicked_articles.items():
+            histories[account_id] = user_clicks
 
         return histories
--- a/src/poprox_storage/repositories/clicks.py
+++ b/src/poprox_storage/repositories/clicks.py
@@ -35,13 +35,14 @@ class DbClicksRepository(DatabaseRepository):
         super().__init__(connection)
         self.tables = self._load_tables("clicks")
 
-    def track_click_in_database(self, newsletter_id, account_id, article_id):
+    def track_click_in_database(self, newsletter_id, account_id, article_id, created_at):
         click_table = self.tables["clicks"]
         with self.conn.begin():
             stmt = insert(click_table).values(
                 newsletter_id=newsletter_id,
                 account_id=account_id,
                 article_id=article_id,
+                created_at=created_at
             )
             self.conn.execute(stmt)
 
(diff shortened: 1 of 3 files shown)


tests/test_clicks.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def test_get_active_survey(pg_url: str):
+    engine = create_engine(pg_url)
+    with engine.connect() as conn:
+        conn.execute(text("delete from clicks;"))
+        conn.execute(text("delete from newsletters;"))
+        conn.execute(text("delete from articles;"))
+
+        dummy_click_1 = uuid4()
```
karlhigley, line 32: We know these are for testing but it's not obvious in the code below that these are the ids and not the click objects (for example), so I'd probably name these something that includes `id` like:
```
click_id_1 = ...
user_id_1 = ...
article_id_1 = ...
```

tests/test_clicks.py
```diff
@@ -0,0 +1,115 @@
...
+        dummy_user_1 = uuid4()
+        dummy_user_2 = uuid4()
+        dummy_newsletter = uuid4()
+
+        dummy_article_1 = uuid4()
+        dummy_article_2 = uuid4()
+
+        query_0 = text(
+                f'''insert into accounts(account_id, email, status, source) values (:account_id, 'user-1@gmail.com', '', '');'''
+            )
```
karlhigley, line 43: `query_0` looks a lot like `DbAccountRepository.create_new_account()`. Could you use that here?

tests/test_clicks.py
```diff
@@ -0,0 +1,115 @@
...
+        query_1 = text(
+                f'''insert into clicks(click_id, account_id, newsletter_id, article_id, created_at)
+                    values (:click_id_1, :account_id, :newsletter_id, :article_id_1, '2024-06-12 09:55:22');'''
+            )
+        query_2 = text(
+                f'''insert into clicks(click_id, account_id, newsletter_id, article_id, created_at)
+                    values (:click_id_2, :account_id, :newsletter_id, :article_id_2,  '2024-07-14 12:55:22');'''
+            )
```
karlhigley, line 51: `query_1` and `query_2` look like `DbClicksRepository.track_click_in_database()`. I see that this is setting the `created_at` field, which that method doesn't yet provide a way to do. Could you add an optional `created_at` param to that method and then use it here?

tests/test_clicks.py
```diff
@@ -0,0 +1,115 @@
...
+        query_3 = text(
+                f'''insert into newsletters(newsletter_id, account_id, content, created_at, html) 
+                values (:newsletter_id, :account_id, '{json.dumps({})}', '{datetime.now()}', '');'''
+            )
```
karlhigley, line 55: `query_3` seems similar to `DbNewsletterRepository.log_newsletter_content()`. Would that work here?

tests/test_clicks.py
```diff
@@ -0,0 +1,115 @@
...
+        query_4 = text(
+            f'''insert into articles(article_id, published_at, created_at, title, content, url) 
+                values (:article_id, '{datetime.now()}', '{datetime.now()}', :title, '{json.dumps({})}', :url);'''
+        )
```
karlhigley, line 59: `query_4` looks like `DbArticleRepository.insert_article()`, which could save you some SQL

karlhigley (commented):
This looks like it's heading in the right direction but could still use some polish

tests/test_clicks.py
sophiasun0515: I tested locally, but this does not work as it has a `with self.conn.begin()` line within the `log_newsletter_content()`, which is contradicting our test db setup. I guess a work around is to hard code the sql command here...

tests/test_clicks.py
sophiasun0515: Hmm, this also does not work as it would throw some weird compilation error to me:
`sqlalchemy.exc.CompileError: Unconsumed column names: preview_image_id, external_id, raw_data, source`

And how I call it is simply created an article object:

`article_id_1 = dbArticleRepository.insert_article(Article(title="Article 1"))`

Not sure what might go wrong, but probably some object-to-sql conversion issue?

tests/test_clicks.py
sophiasun0515: Ahh, found the same complaint about double db connection here as well 😅

tests/test_clicks.py
karlhigley: > Ah I see, but looks like create_new_account() function does not take in account_id as a param

If you needed that, you could make it an optional parameter to that function with a default value of `None` to maintain backward compatibility for existing callers. None of the repository code is set in stone or even particularly good (which I can say with confidence because I wrote most of it), so you should feel free to make changes where you need to. However...

> also it returns an Account object 🤔 I can see it might be able to replace the two account objects created in line 103, but again does that mean we might need to manually fetch the created Account object and its account_id for other following queries?

As the code is currently written, you need to know the account ids up front, then use them to create the database state you need, then create corresponding account objects—which works fine but is kind of a lot of work. If you use `create_new_account()` instead somewhere up here in the setup code (before line 100 or so), then:
* You wouldn't need to manually generate the account ids with `uuid4()` because they'll be auto-assigned by `create_new_account()` (so you probably don't actually need to pass the ids in)
* You can assign the result to `accounts` (which would replace lines 103-104, as you point out)
* You can pull the auto-assigned account ids from the `Account` objects with `accounts[0].account_id`/`accounts[1].account_id` when you need to use them to record clicks below

It'll do essentially the same thing, just at a higher level of abstraction so you won't have to think as hard or write so much SQL/code.

tests/test_clicks.py
karlhigley: Not opposed to changing the code upstream to make it easier to use, but...what about that is causing problems? What issues or errors do you see? Something about transactions, I'm guessing?

tests/test_clicks.py
karlhigley: That is a weird one! That's worth digging into to figure out what's going on; let's do that and sort it out so that you can use that here.

karlhigley (commented):
It's a little hard to tell why you're encountering issues without seeing the current state of the code. Could you commit and push the state of the code you're working from (even if it's not finished)?

sophiasun0515:
Notes from debugging with Karl:

If weird db complain happen in test, when running `source dev/init_poprox_dev.sh`, make sure it doesn't throw any error and produce the log containing latest changes recorded in `poprox-db/migrations/version`.

src/poprox_storage/repositories/clicks.py
```diff
@@ -35,13 +35,14 @@ def __init__(self, connection):
         super().__init__(connection)
         self.tables = self._load_tables("clicks")
 
-    def track_click_in_database(self, newsletter_id, account_id, article_id):
+    def track_click_in_database(self, newsletter_id, account_id, article_id, created_at):
```
karlhigley, line 38: In order to avoid breaking the code that calls this function, `created_at` needs a default so that it can be provided but isn't mandatory:
```suggestion
    def track_click_in_database(self, newsletter_id, account_id, article_id, created_at=None):
```

karlhigley (approved):
One minor suggestion, but otherwise looks good!

src/poprox_storage/repositories/clicks.py
sophiasun0515: Sure, thanks for the suggestion!

PR #18, opened 2024-07-17
Add instructions for running the database to the README

karlhigley:


kluver (approved):
This is accurate information.

I'm not sure if this is how we actually want tests working -- but it is an accurate description of how they work right now.

PR #20, opened 2024-07-19
Standardize code style guidelines and pre-commit hooks.

kluver:


kluver:
arrgh more tests failing. ok, well I'll get back on this.

kluver:
I apparently never did `pre-commit install` I'm thinking that's why I kept having conflicts. We should be ready to merge. Sorry for the false-notice.

karlhigley (approved):

PR #22, opened 2024-07-26
Add `opt_out` column to `expt_assignments` table

karlhigley:
This also renames the table from `expt_allocations` to `expt_assignments`. That felt more urgent now that we're going to have additional tables (like `datasets` and `account_aliases`) that will connect to related tables like `accounts`.

### DIFF
--- a/poprox-db/migrations/versions/2024_07_26_0945-fc8d79481ee4_merge_head_revisions.py
+++ b/poprox-db/migrations/versions/2024_07_26_0945-fc8d79481ee4_merge_head_revisions.py
@@ -5,15 +5,12 @@ Revises: 14be38931ae0, 5d566fff8aaa
 Create Date: 2024-07-26 09:45:33.246538
 
 """
+
 from typing import Sequence, Union
 
-from alembic import op
-import sqlalchemy as sa
-
-
 # revision identifiers, used by Alembic.
-revision: str = 'fc8d79481ee4'
-down_revision: Union[str, None] = ('14be38931ae0', '5d566fff8aaa')
+revision: str = "fc8d79481ee4"
+down_revision: Union[str, None] = ("14be38931ae0", "5d566fff8aaa")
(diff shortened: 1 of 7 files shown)


poprox-db/migrations/versions/2024_07_26_0945-fc8d79481ee4_merge_head_revisions.py
```diff
@@ -0,0 +1,26 @@
+"""merge head revisions
```
kluver, line 1: Yeah, I manually made the pre-commit fix revisions in the past, but I think there is an argument that we should maybe just exclude these?

kluver: I have code that ignores the `poprox-db/migrations/versions` folder when doing code style checks. I'm holding back committing pending your thoughts on the topic.

karlhigley: No strong opinion either way, but I went ahead and applied the formatting/fixes from the linters

kluver: The default template has code style errors in it (unused imports I think) so I think it might be worth it. If you have no strong opinions I think I'm going to go ahead and do it.

kluver:
I was hoping to downstream some changes from this PR (so I didn't have to deal with the alembic merge) -- I don't think this passes the database tests.

kluver:
> I was hoping to downstream some changes from this PR (so I didn't have to deal with the alembic merge) -- I don't think this passes the database tests.

```
    if isinstance(revision.down_revision, tuple):
        return  # don't test merge revisions
```
Adding this to the staircase test "fixes" it -- it does prevent testing merge revisions, but I think we can plan to always have empty merge revisions

PR #26, opened 2024-07-29
add header tracking column to click tracking to let us investigate non-click link visits

kluver:


karlhigley (approved):

karlhigley:
This may need a merge revision in order to avoid a "multiple revision heads" situation. We don't have a test for that yet, but I found something about how to write one, so maybe we can add it soon.

kluver:
Sounds good. I just had a few thoughts "click" into place so I'm going to give it a good 20 minute try at actually testing this code locally before we deploy as well.

PR #27, opened 2024-07-31
Fetch Newsletter Function With Unit Test

sophiasun0515:
Added for newsletter export. Unit test passed locally.

### DIFF
--- a/src/poprox_storage/repositories/newsletters.py
+++ b/src/poprox_storage/repositories/newsletters.py
@@ -1,10 +1,16 @@
+import json
+from datetime import datetime
 from sqlalchemy import (
     Connection,
     Table,
     insert,
+    select,
 )
 
+from uuid import UUID
+from collections import defaultdict
 from poprox_storage.repositories.data_stores.db import DatabaseRepository
+from poprox_concepts import Account, Article
 
 
 class DbNewsletterRepository(DatabaseRepository):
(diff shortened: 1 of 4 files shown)


src/poprox_storage/repositories/newsletters.py
```diff
@@ -20,21 +26,54 @@ def store_newsletter(self, newsletter_id, account_id, recommended_articles, arti
         newsletter_table = self.tables["newsletters"]
         impression_table = self.tables["impressions"]
 
-        with self.conn.begin():
-            stmt = insert(newsletter_table).values(
-                newsletter_id=newsletter_id,
-                account_id=str(account_id),
-                content=[rec.json() for rec in recommended_articles],
-                html=article_html,
```
sophiasun0515, line 28: I removed the `with self.conn.begin():` line since it will raise SQLAlchemy request error in the new unit test (`dbNewsletterRepository.log_newsletter_content(newsletter_1_id, user_account_1.account_id, newsletter_1_articles, "fake-url-1")`) with the following message:

> E           sqlalchemy.exc.InvalidRequestError: This connection has already initialized a SQLAlchemy Transaction() object via begin() or autobegin; can't call begin() here unless rollback() or commit() is called first. 

But the `test_clicks.py` does not have this issue with the same call of `log_newsletter_content()`, which is pretty weird 😅 

Can this removal of `self.conn.begiin()` impact anything?

karlhigley: Yeah, this is a result of SQLalchemy's automatic transaction management, which is very convenient until it isn't. `with self.conn.begin()` tells SQLalchemy that everything inside the `with` block should be executed in a single database transaction, so that it either all succeeds or all fails together. This avoids creating a fragment of a newsletter in the database if something in the middle fails.

The issue that the error message is describing is that there's already a transaction in progress so it can't start another one, and the solution to that is in the last part of the error: `can't call begin() here unless rollback() or commit() is called first`. `commit()` says "okay, I'm finished with this set of changes, so go ahead and apply them to the database for real now", and `rollback()` says "oops, I messed up, so erase any database changes since `commit()` was last called."

Instead of removing the `with self.conn.begin()` you can call `self.conn.commit()` before `log_newsletter_content()` in the test.

sophiasun0515: Sure, thanks for the information!

karlhigley (approved):

kluver (approved):

PR #28, opened 2024-08-06
Fix Newsletter Export Data Type Complaint with Json Dumps

sophiasun0515:
While testing the PR of poprox-platform: https://github.com/CCRI-POPROX/poprox-platform/pull/131, we encountered an export error that would complain about json export; with this data type check we can get rid of it.

### DIFF
--- a/src/poprox_storage/repositories/newsletters.py
+++ b/src/poprox_storage/repositories/newsletters.py
@@ -59,6 +59,8 @@ class DbNewsletterRepository(DatabaseRepository):
         for row in newsletter_result:
             raw_articles = []
             for article_json in row.content:
+                if isinstance(article_json, dict):
+                    article_json = json.dump(article_json)
                 raw_articles.append(json.loads(article_json))
             articles = [
                 Article(
--- a/src/poprox_storage/repositories/newsletters.py
+++ b/src/poprox_storage/repositories/newsletters.py
@@ -60,7 +60,7 @@ class DbNewsletterRepository(DatabaseRepository):
             raw_articles = []
             for article_json in row.content:
                 if isinstance(article_json, dict):
-                    article_json = json.dump(article_json)
+                    article_json = json.dumps(article_json)
                 raw_articles.append(json.loads(article_json))
             articles = [
                 Article(
--- a/src/poprox_storage/repositories/newsletters.py
+++ b/src/poprox_storage/repositories/newsletters.py
@@ -64,11 +64,12 @@ class DbNewsletterRepository(DatabaseRepository):
                 raw_articles.append(json.loads(article_json))
             articles = [
                 Article(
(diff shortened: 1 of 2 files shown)


src/poprox_storage/repositories/newsletters.py
```diff
@@ -59,14 +59,17 @@ def fetch_newsletters(self, accounts: list[Account]) -> dict[UUID, dict[UUID, li
...
             for article_json in row.content:
+                if isinstance(article_json, dict):
+                    article_json = json.dumps(article_json)
                 raw_articles.append(json.loads(article_json))
             articles = [
                 Article(
+                    article_id=raw["article_id"],
                     title=raw["title"],
-                    content=raw.get("description", None),
+                    content=raw.get("content", None),
```
karlhigley, line 69: I'm not sure about this line—I know it looked weird but my recollection is that this was an artifact of different terminology between us and AP. What we call the `content` is really the blurb/slug and I don't think they call it that. If this is working the way you expect then it's probably okay, but just want to double-check this part.

sophiasun0515: Ah gotcha! I'm applying this change since if we don't change it, the new line in the test would fail and return none:

`assert "article content 1" == user_1_newsletter[newsletter_1_id][0].content`

Is that expected or am I calling the wrong field in the test? I've printed out the Article object and looks like it just has `content` field instead of `description`.

karlhigley: For the short term, I think that line should be:
```
assert "article content 1" == user_1_newsletter[newsletter_1_id][0].description
```

Outside the scope of this PR, we should probably clarify the names of the fields in the `Article` class to make this easier to figure out

sophiasun0515: Hmm this is weird, since if I change it to `description`, we'll see such error:

`tests/test_newsletters.py::test_fetch_newsletters - AttributeError: 'Article' object has no attribute 'description'`

Referring to the [article object definition](https://github.com/CCRI-POPROX/poprox-concepts/blob/c0552428f046dcb1f706f421b3ae637e8022d9a6/src/poprox_concepts/domain/article.py#L28), indeed that's what this error message is complaining about....do I miss anything here?

karlhigley: Maybe I'm misunderstanding which data this is operating on! I thought it was comparing Article objects to AP JSON, but I'm realizing now maybe it's comparing Article objects to serialized versions of the same objects?

sophiasun0515: Yeah I think it should be the same object serialization case as you mentioned. So the current setup using `content` in all field to construct the `Article` object can pass, while the `description` would fail. If the current setup looks good I think we're ready to merge it!

PR #30, opened 2024-08-12
Only fetch the latest version of each article

karlhigley:
This will help us avoid duplicate articles in the newsletter caused by updates from AP's side that they provide to us as separate articles in the feed.

### DIFF
--- a/src/poprox_storage/repositories/articles.py
+++ b/src/poprox_storage/repositories/articles.py
@@ -6,6 +6,9 @@ from uuid import UUID
 import boto3
 from sqlalchemy import (
     Connection,
+    Table,
+    and_,
+    func,
     select,
 )
 from tqdm import tqdm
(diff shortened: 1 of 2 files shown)


src/poprox_storage/repositories/articles.py
```diff
@@ -170,8 +173,28 @@ def store_mention(self, mention: Mention) -> UUID | None:
     insert_entity = store_entity
     insert_mention = store_mention
 
-    def _get_articles(self, article_table, where_clause=None) -> list[Article]:
-        query = article_table.select()
+    def _get_articles(self, article_table: Table, where_clause=None) -> list[Article]:
+        # Select only the most recent article row for each source/external id pair
+        inner_query = (
```
kluver, line 178: This might be pre-mature optimization -- but do we want to put the where clause into the inner query instead of leaving it in the outer query?

kluver (approved):

src/poprox_storage/repositories/articles.py
karlhigley: I thought about that too! The thing that prevented me from doing so is that the inner query doesn't fetch all the fields so can't support arbitrary where clauses on the table. There may be a way to optimize this, but it wasn't immediately obvious to me how

PR #31, opened 2024-08-13
Remove repository method name aliases

karlhigley:
We put these in for backward compatibility while we updated the other repos which will soon be complete, so we're ready to remove the aliases.

### DIFF
--- a/src/poprox_storage/repositories/accounts.py
+++ b/src/poprox_storage/repositories/accounts.py
@@ -67,7 +67,7 @@ class DbAccountRepository(DatabaseRepository):
 
     def fetch_unassigned_accounts(self, start_date: date, end_date: date):
         account_tbl = self.tables["accounts"]
-        allocation_tbl = self.tables["expt_allocations"]
+        assignment_tbl = self.tables["expt_assignments"]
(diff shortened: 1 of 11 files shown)


src/poprox_storage/repositories/accounts.py
```diff
@@ -93,11 +93,11 @@ def fetch_unassigned_accounts(self, start_date: date, end_date: date):
         group_ids = self._id_query(group_query)
 
         # Find the users allocated to those groups
-        allocation_query = select(allocation_tbl.c.account_id).where(allocation_tbl.c.group_id.in_(group_ids))
-        allocation_account_ids = self._id_query(allocation_query)
+        assignment_query = select(assignment_tbl.c.account_id).where(assignment_tbl.c.group_id.in_(group_ids))
```
kluver, line 96: I'm pretty sure we made this change a while ago. Maybe we're just not using this function?

kluver (approved):

src/poprox_storage/repositories/accounts.py
karlhigley: Yeah, that's exactly it—we haven't tried to do anything with experiment assignments since the change happened

PR #33, opened 2024-08-16
Create `ArticlePlacement` concept and corresponding table

JUNYUANGO:
`test_scrape_article.py`
Here is the scraping code. It can now successfully scrape about 50 articles after each run. They can be stored in a CSV file as well as the local PostgreSQL server. 

`test_create_scrape_table.py`
Here is the code for creating a new table. 

`apnews_articles.csv`
A slight change to the columns of scraped data. Examples are shown here. 

This is the testing code. Will work with Ishan to build further features.


kluver:
One question before I take a look at the code: Is this PR a draft? It sounds like you're looking for work-in-progress feedback rather than saying this code is ready for deployment.

Also -- don't sweat the linter failures, we can fix that later. You can avoid them by installing the pre-commit hook that will run the format checkers before commit. I think the command for that is `pre-commit install`

.gitignore
```diff
@@ -160,4 +160,4 @@ cython_debug/
 #  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
-#.idea/
```
kluver, line 163: Any particular reason for this?

dev/scraped-articles/apnews_articles_2024-08-16_14-22-20.csv
```diff
@@ -0,0 +1,50 @@
+title,description,url,section,level,image_url,scraped_at
```
kluver, line 1: we _probably_ don't want to store these article csv's in github. Is there a compelling reason to keep these files or are these just temporary?

poprox-db/tests/test_create_scrape_table.py
```diff
@@ -0,0 +1,59 @@
+import psycopg2
+from sqlalchemy import create_engine
+
+# Database connection parameters
+conn_params = {
```
kluver, line 5: This is not how the rest of the codebase is connecting -- so we'll need to clean this part up before merging.

poprox-db/tests/test_create_scrape_table.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+# Function to test the creation of a new table and insertion of data
+def test_create_scrape_table():
+    # Connect to the PostgreSQL server
+    conn = psycopg2.connect(**conn_params)
+    cur = conn.cursor()
+
+    # Create a new table for testing
+    cur.execute("""
```
kluver, line 26: This is not how we're managing create table commands we will need to clean this up.

To see how we're managing this type of change I recommend looking at the files in `poprox-storage/poprox-db/migrations/versions`

dev/test_scrape_article.py
```diff
@@ -0,0 +1,264 @@
...
+
+import pandas as pd
+import psycopg2
+import requests
+from bs4 import BeautifulSoup
+
+
+# Function to request and parse via BeautifulSoup
+def request_and_parse(url):
+    time.sleep(random.random())  # Add a random waiting time (less than 1 sec) for each request
```
kluver, line 13: We might have a library for managing the timeout -- I would have to look more into this later.

kluver (commented):
The parsing code itself looks fine. How we integrate it with poprox will need some cleaning up to match how we do other work in poprox (also the parsing code doesn't really belong in this repo in the first place -- poprox code is split over multiple repositories and only the code for storing and fetching data belongs here -- the "behavior" normally would go somewhere else.

I think the next step might be finding another time to work together or do some more in-depth Q&A over slack. We've got the right code, now we've got to get it properly integrated with the rest of the project.

poprox-db/tests/test_create_scrape_table.py
JUNYUANGO: Roger, I will clean up them later.

kluver: Let me know if you want to pair up on that.

.gitignore
JUNYUANGO: Emm, I use PyCharm to write all the codes and try to ignore this folder which is created by JetBrains.

JUNYUANGO:
Hi Daniel, here is what I have done so far, please check the last two commits:
1. Create a new version to have a table for scraped articles.
2. Move the test code to another folder, rewrite codes inside to test connection and sql stuffs.

karlhigley:
This is looking better! I made a few more small changes to remove the CSV files and move the scraping code into `src/poprox_storage`. Some of the code—in particular the part that interacts with the database—belongs there, but the scraping itself should move to `poprox-platform`. Let's see if we can find a way to split the two pieces apart from each other here in this PR and then we can move the scraping part over there once we've done that.

src/poprox_storage/scrape_articles.py
```diff
@@ -0,0 +1,264 @@
...
+now = datetime.datetime.now()
+now = now.strftime("%Y-%m-%d_%H-%M-%S")
+
+# # Save to the folder "articles"
+# path_to_save = f'scraped-articles/apnews_articles_{now}.csv'
+# df.to_csv(path_to_save, index=False)
+# print(f'Articles saved to {path_to_save}')
+
+# Database Connection and Table Creation
+
```
karlhigley, line 169: Everything from here down looks like it's inserting data in the database, but doing so very manually. You can write raw SQL like this and it will work okay, but it's hard to re-use and maintain. This library (`poprox-storage`) provides repository classes (like [this one for user clicks on articles](https://github.com/CCRI-POPROX/poprox-storage/blob/main/src/poprox_storage/repositories/clicks.py#L32)) that provide a plain Python interface for other parts of the POPROX code base to use when interacting with the database (and other storage like S3.)

Let's see if we can reshape this code in that direction. (More on that below.)

src/poprox_storage/scrape_articles.py
```diff
@@ -0,0 +1,264 @@
...
+    "user": "postgres",
+    "password": "thisisapoproxpassword",
+    "host": "127.0.0.1",
+    "port": "5433",
+}
+
+# Connect to the PostgreSQL server
+try:
+    conn = psycopg2.connect(**conn_params)
+    print("Connection successful!")
```
karlhigley, line 182: Repository classes take a database connection [as a constructor param](https://github.com/CCRI-POPROX/poprox-storage/blob/main/src/poprox_storage/repositories/clicks.py#L32), so you won't have to create one yourself. That lets us share database connections across the code that interacts with different parts of the database, which minimizes the number of connections we need to have open.

src/poprox_storage/scrape_articles.py
```diff
@@ -0,0 +1,264 @@
...
+    cur.execute("""
+        SELECT EXISTS (
+            SELECT 1
+            FROM information_schema.tables
+            WHERE table_schema = 'public'
+            AND table_name = 'scraped_articles'
+        );
+    """)
+
+    table_exists = cur.fetchone()[0]
```
karlhigley, line 197: Since you wrote a database migration, you can assume that the table will exist and don't need to manually check that.

src/poprox_storage/scrape_articles.py
```diff
@@ -0,0 +1,264 @@
...
+                url VARCHAR PRIMARY KEY,
+                section VARCHAR,
+                level VARCHAR,
+                image_url VARCHAR,
+                scraped_at TIMESTAMP
+            );
+        """)
+        print("Table 'scraped_articles' created successfully!")
+    else:
+        print("Table 'scraped_articles' already exists.")
```
karlhigley, line 214: Similarly, creating the table is already handled by the migration so this code doesn't need to (and shouldn't) try to create the table on its own

src/poprox_storage/scrape_articles.py
```diff
@@ -0,0 +1,264 @@
...
+        WHERE table_name = 'scraped_articles';
+    """)
+
+    # Fetch all the results
+    schema = cur.fetchall()
+
+    # Print the schema of the table
+    print("Schema of the 'scraped_articles' table:")
+    for column in schema:
+        print(f"Column: {column[0]}, Data Type: {column[1]}")
```
karlhigley, line 229: We know what the schema will be since it's specified in the migration, so we can skip printing it out here

src/poprox_storage/scrape_articles.py
```diff
@@ -0,0 +1,264 @@
...
+                    row["url"],
+                    row["section"],
+                    row["level"],
+                    row["image_url"],
+                    row["scraped_at"],
+                ),
+            )
+            print(f"Inserted row {idx + 1} into 'scraped_articles' table.")
+        except Exception as e:
+            print(f"Error inserting row {idx + 1}: {e}")
```
karlhigley, line 252: This part looks like it could become a `store_scraped_article` method on a new `DbScrapedArticleRepository` class. Here's [an example](https://github.com/CCRI-POPROX/poprox-storage/blob/main/src/poprox_storage/repositories/clicks.py#L37-L48) of a `store` method that does something similar using `sqlalchemy`'s API (instead of raw SQL) that you can use as a template.

tests/test_create_scrape_table.py
```diff
@@ -0,0 +1,61 @@
...
+
+@pytest.fixture(scope="session")
+def pg_url():
+    """
+    Provides base PostgreSQL URL for creating temporary databases.
+    """
+    return os.getenv("CI_POPROX_PG_URL", DEFAULT_PG_URL)
+
+
+def test_scrape_table(pg_url: str):
```
karlhigley, line 21: A test should run some part of the code from outside the tests and assert on the results that come back from that code. As currently written, this is mostly testing that SQL works as expected rather than testing that your code above works as expected because it isn't calling any of the code above.

Once we have a `DbScrapedArticleRepository` class with a `store_scraped_article` method, this test can call that method to insert a row and then we can check that the row was actually inserted into the table (as you do below.)

karlhigley (commented):
You have a lot of the right pieces here, but there's work to do in making it match the structure and style of the rest of our code. I've left comments pointing out where and how I'd refactor this—it's fine if you want to take a stab at that and communicate through comments here, but I'm also happy to get on a call and work on this together in a VSCode LiveShare session.

JUNYUANGO:
To sum up, this is what we're at right now:
1. Migration code is ready.
2. DbScrapeRepository is created with three basic functions (store, fetch by url, delete), able to access the DB. 

One issue here is whether we use a different class for scraped article. [(See PR here)](https://github.com/CCRI-POPROX/poprox-concepts/pull/23).

JUNYUANGO:
One confusion here is: I tried `ruff check --fix` to auto-fix the reported error by ruff, but this would fail the pre-commit tests. If I ran the  `pre-commit run --all-files`, this would fail the `ruff check .`. Which one should I keep, or is there the correct way I missed?

tests/test_create_placement_table.py
```diff
@@ -0,0 +1,54 @@
...
+def pg_url():
+    """
+    Provides base PostgreSQL URL for creating temporary databases.
+    """
+    return os.getenv("CI_POPROX_PG_URL", DEFAULT_PG_URL)
+
+
+def test_placement_table(pg_url: str):
+    engine = create_engine(pg_url)
+    with engine.connect() as conn:
```
karlhigley, line 26: We do database setup in a lot of tests, so there's already [a test fixture](https://github.com/CCRI-POPROX/poprox-storage/blob/main/tests/conftest.py#L22) for this that will simplify the code here:
```suggestion
def test_placement_table(db_engine):
    with db_engine.connect() as conn:
```

karlhigley (commented):
Looking good! Made a few small changes and left some minor comments to address, but I think this is nearly ready to merge.

JUNYUANGO:
Hi Karl, this code should now be working as you expect.

karlhigley:
@JUNYUANGO Looks good! The next step from here is to set up the scraping and linking code over in [poprox-platform](https://github.com/CCRI-POPROX/poprox-platform). The details will be different (since you're pulling data from a different place), but you can use [this function](https://github.com/CCRI-POPROX/poprox-platform/blob/main/poprox_platform/ingest/ap.py#L24-L65) that pulls data from the API as a template for the scraping part. That part should scrape pages and write data to S3, and then I'm imagining a second function that loads the scraped data, finds the corresponding articles, and creates placements for them.

PR #35, opened 2024-08-19
Auto-register `DatabaseRepository` sub-classes to create with decorator

karlhigley:
This is the deep dark magic that allows auto-detection of which repository classes are available to be constructed by the decorator.

### DIFF
--- a/src/poprox_storage/repositories/data_stores/db.py
+++ b/src/poprox_storage/repositories/data_stores/db.py
@@ -48,7 +48,6 @@ class DatabaseRepository:
         Gets called once for each loaded class that sub-classes DatabaseRepository
         """
 
-        # Register the attribute name and corresponding class
         cls._repository_types.add(cls)
 
     def _load_tables(self, *args) -> dict[str, Table]:
--- a/src/poprox_storage/repositories/data_stores/db.py
+++ b/src/poprox_storage/repositories/data_stores/db.py
@@ -18,18 +18,12 @@ logger = logging.getLogger(__name__)
 logger.setLevel(logging.DEBUG)
 
 
-DbRepositories: type[BaseModel] = create_model(
-    "DbRepositories",
-    conn=(Connection, None),
-    __config__=ConfigDict(arbitrary_types_allowed=True),
-)
-
-
-def camel_to_snake(s):
-    return "".join(["_" + c.lower() if c.isupper() else c for c in s]).lstrip("_")
+DbRepositories: type[BaseModel] | None = None
 
 
 def db_repositories(*args):
+    create_repository_model()
+
     if len(args) == 1 and not isinstance(args[0], list):
         names = [args[0]]
     else:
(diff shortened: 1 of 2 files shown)


src/poprox_storage/repositories/data_stores/db.py
```diff
@@ -16,10 +18,67 @@
 logger.setLevel(logging.DEBUG)
 
 
+DbRepositories: type[BaseModel] = create_model(
+    "DbRepositories",
+    conn=(Connection, None),
+    __config__=ConfigDict(arbitrary_types_allowed=True),
+)
```
karlhigley, line 25: This dynamically constructs a Pydantic model with a field for the database connection, which provides a default value here in case we don't detect any sub-classes of `DatabaseRepository` and establishes `DbRepositories` as a thing before we get to the decorator below

src/poprox_storage/repositories/data_stores/db.py
```diff
@@ -16,10 +18,67 @@
...
+        @wraps(handler)
+        def wrapper(event, context):
+            type_hints = get_type_hints(DbRepositories)
+            with DB_ENGINE.connect() as conn:
+                # create Repositories object and pre-construct all requested repository objects
+                repos = DbRepositories(conn=conn)
+                for name in names:
+                    repo_type = type_hints[name]
+                    setattr(repos, name, repo_type(conn))
+                return handler(event, context, repos)
```
karlhigley, line 48: The magic here gets a bit simpler because we no longer have explicitly annotated `DbSomethingRepository | None` types, so we don't have to extract type arguments from union types

src/poprox_storage/repositories/data_stores/db.py
```diff
@@ -16,10 +18,67 @@
...
         self.conn: Connection = connection
 
+    def __init_subclass__(cls, *args, **kwargs):
+        """
+        Gets called once for each loaded class that sub-classes DatabaseRepository
+        """
+        global DbRepositories
+
+        # Register the attribute name and corresponding class
+        concept_name = camel_to_snake(cls.__name__.removeprefix("Db").removesuffix("Repository"))
```
karlhigley, line 68: This relies on a naming convention, which we may or may not want to do. An alternate way to go about this would be to put a property or something on each class that defines what string this should be.

src/poprox_storage/repositories/data_stores/db.py
```diff
@@ -16,10 +18,67 @@
...
+        # Convert the attribute names and classes to fields
+        fields = {name: Annotated[type_, Field(default=None)] for name, type_ in cls._repository_types.items()}
+
+        # Dynamically create a Pydantic model with those fields
+        DbRepositories = create_model(
+            "DbRepositories",
+            conn=(Connection, None),
+            **fields,
+            __config__=ConfigDict(arbitrary_types_allowed=True),
+        )
```
karlhigley, line 80: Every time we find another sub-class we create a new model type. That's not ideal but we don't know that we're done on any particular invocation of this hook. It might be possible to defer this until the decorator gets called though.

karlhigley:
A downside of this approach is that it makes type annotations hard again. In order to get that working, I'm thinking about requiring an argument on the handler function for each repository (which could then be annotated.)

kluver (approved):

src/poprox_storage/repositories/data_stores/db.py
```diff
@@ -16,10 +17,40 @@
...
+
     def __init__(self, connection: Connection):
         self.conn: Connection = connection
 
+    def __init_subclass__(cls, *args, **kwargs):
+        """
+        Gets called once for each loaded class that sub-classes DatabaseRepository
+        """
+
+        # Register the attribute name and corresponding class
```
karlhigley, line 51:
```suggestion
```

PR #36, opened 2024-08-20
Allow fetching article by ingestion date instead of publication date

karlhigley:
This will make it easier to send emails immediately after on-boarding, because we won't have the "sliding window" effect where articles slide beyond one day ago from the current time, gradually reducing the size of the candidate pool throughout the day until we ingest the next day's new articles.

Depended on by https://github.com/CCRI-POPROX/poprox-platform/pull/139

### DIFF
--- a/src/poprox_storage/repositories/articles.py
+++ b/src/poprox_storage/repositories/articles.py
@@ -34,24 +34,36 @@ class DbArticleRepository(DatabaseRepository):
             "impressions",
         )
 
-    def fetch_todays_articles(self) -> list[Article]:
-        return self.fetch_articles_since(days_ago=1)
-
-    def fetch_past_articles(self) -> list[Article]:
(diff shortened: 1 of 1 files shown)


src/poprox_storage/repositories/articles.py
```diff
@@ -34,24 +34,36 @@ def __init__(self, connection: Connection):
             "impressions",
         )
 
-    def fetch_todays_articles(self) -> list[Article]:
-        return self.fetch_articles_since(days_ago=1)
-
-    def fetch_past_articles(self) -> list[Article]:
-        return self.fetch_articles_before(days_ago=1)
-
```
karlhigley, line 42: There's only one usage of either of these methods, which is used in `poprox-platform` to fetch the candidate pool for today's newsletter. Removed in the corresponding PR that uses the new methods below.

kluver (approved):

PR #37, opened 2024-08-20
Clean up database test fixtures

karlhigley:
In order to work seamlessly with a local database, we also need to use the same database engine that the connection was created from. Otherwise we run the risk of e.g. connecting to the prod database and running a bunch of delete statements.


kluver (approved):
lgtm, 

If the concern is about connecting to the "wrong" database is there some way to take this further? Just brainstorming, but we could probably instrument the test database to put a table with a specific value in it and we can do a positive check for that on test start up and halt-out if we don't see that?

karlhigley:
This was less of an abstract concern and more of a "wait, I can't actually run the tests because it's trying to talk to two different databases" 😅

PR #38, opened 2024-08-20
Switch to using direct repository injection

kluver:


src/poprox_storage/repositories/data_stores/db.py
```diff
@@ -18,58 +17,29 @@
...
-
-
-def camel_to_snake(s):
-    return "".join(["_" + c.lower() if c.isupper() else c for c in s]).lstrip("_")
+    return wrapper
 
 
 class DatabaseRepository:
-    _repository_types = {}
+    _repository_types = set()
```
kluver, line 42: Switched to set since we don't care what the parameter is named anymore -- this does formally require annotating parameters, but working with this type of magic that's probably for the best.

src/poprox_storage/repositories/data_stores/db.py
```diff
@@ -18,58 +17,29 @@
 logger.setLevel(logging.DEBUG)
 
 
-DbRepositories: type[BaseModel] | None = None
+def inject_repos(handler):
+    @wraps(handler)
+    def wrapper(event, context):
+        with DB_ENGINE.connect() as conn:
+            # this will be a dictionary containing all parameters (with type annotations)
```
karlhigley, line 24: This could probably be a type annotation along the lines of `dict[str, type]`

karlhigley (approved):

PR #39, opened 2024-08-20
Apply `Newsletter` domain concept to storing emails

karlhigley:
Connected PRs:
* https://github.com/CCRI-POPROX/poprox-concepts/pull/21
* https://github.com/CCRI-POPROX/poprox-storage/pull/39
* https://github.com/CCRI-POPROX/poprox-platform/pull/140

### DIFF
--- a/src/poprox_storage/repositories/data_stores/db.py
+++ b/src/poprox_storage/repositories/data_stores/db.py
@@ -11,6 +11,8 @@ from sqlalchemy import (
 from sqlalchemy.dialects.postgresql import insert
 from sqlalchemy.exc import IntegrityError, InternalError
 
+from poprox_storage.aws import DB_ENGINE
+
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.DEBUG)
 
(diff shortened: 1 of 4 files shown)


sophiasun0515 (approved):
Looks good! I was thinking if we should also update the `fetch_newsletters()` call to use the new object, but after reading it again maybe the current logic still works?

karlhigley:
We should! I didn't do it in this PR for two reasons:
* The motivation for this PR came from working toward restructuring the platform code that generates newsletter emails
* I expect to revisit the newsletter fetching code soon in the context of exporting complete anonymized datasets (across newsletters, clicks, etc)

sophiasun0515:
> We should! I didn't do it in this PR for two reasons:
> 
> * The motivation for this PR came from working toward restructuring the platform code that generates newsletter emails
> * I expect to revisit the newsletter fetching code soon in the context of exporting complete anonymized datasets (across newsletters, clicks, etc)

Sgtm! Let's ship this one first.

kluver (approved):

PR #41, opened 2024-09-03
Create teams, datasets, and account aliases when ingesting manifests

karlhigley:


karlhigley:
This does create e.g. account aliases, but the methods for fetching them don't exist yet so it doesn't explicitly test that. I'm envisioning that as a separate set of PRs that integrates the new aliases into the recommendation request and dataset export processes.

src/poprox_storage/repositories/accounts.py
```diff
@@ -55,6 +55,21 @@ def fetch_account_by_email(self, email: str) -> Account | None:
             return accounts[0]
         return None
 
+    def store_account(self, account: Account) -> UUID | None:
```
kluver, line 58: Is it worth revising the `web` code that uses the following `store_new_account` to use this version instead?

kluver, line 58: To match existing semantics, should this insert or update?

poprox-db/migrations/versions/2024_09_04_1125-8bf414e0ddfb_remove_name_column_from_datasets_table.py
```diff
@@ -0,0 +1,29 @@
...
+
+# revision identifiers, used by Alembic.
+revision: str = "8bf414e0ddfb"
+down_revision: Union[str, None] = "dd50d8e7777e"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.drop_column("datasets", "dataset_name")
```
kluver, line 22: I don't have a problem with this change, but it does feel unconnected from the other changes, and I'm not immediately seeing the reasoning here.

src/poprox_storage/concepts/manifest.py
```diff
@@ -85,34 +94,43 @@ def manifest_to_experiment(manifest: ManifestFile) -> Experiment:
     start_date = manifest.experiment.start_date or (date.today() + timedelta(days=1))  # noqa: DTZ011
     end_date = start_date + convert_duration(manifest.experiment.duration)
 
+    owner = Team(
+        team_id=manifest.owner.team_id,
+        team_name=manifest.owner.team_name,
+        members=manifest.owner.members,
```
kluver, line 100: Is the implication here that the manifest will contain a list of member UUIDs? Do we like this, or should we have the manifest list emails and then look up the UUIDs as part of ingress here?

src/poprox_storage/paths.py
```diff
@@ -0,0 +1,61 @@
...
+@overload
+def project_root(*, require: bool) -> Path | None: ...
+def project_root(*, require: bool = True) -> Path | None:
+    """
+    Find the project root directory (when we are running in the project).
+
+    This searches upwards from the **current working directory** to find the
+    root of the project, which it identifies by the ``pyproject.toml`` file.  If
+    this function is called from a directory that is not within a checkout of
+    the ``poprox-recommender`` repository, it will raise an error.
```
kluver, line 23: I _think_ the reference to `poprox-recommender` here is a copy-paste typo?

src/poprox_storage/concepts/manifest.py
kluver: Also -- from a type-safety standpoint (which might be uninteresting FWIW) would this be `list[str]` or `list[UUID]` at this point? I feel like it would be `list[str]` but the `Team` class has `list[UUID]`?

Not sure we care about that, but it occurred to me that it's worth checking.

tests/concepts/test_manifest.py
```diff
@@ -0,0 +1,25 @@
+from uuid import UUID
+
+from poprox_storage.concepts.experiment import Experiment, Team
+from poprox_storage.concepts.manifest import manifest_to_experiment, parse_manifest_toml
+from poprox_storage.paths import project_root
+
+
+def test_load_manifest():
+    with open(project_root() / "tests" / "data" / "sample_manifest.toml") as f:
```
kluver, line 9: ` str / str` is new syntax to me. Does it just put in OS appropriate slashes?

kluver (approved):
A few things I want to check, but nothing that looked like a blocker to me.

src/poprox_storage/repositories/accounts.py
karlhigley: I'm not totally sure, but maybe? I left the other method there so this wouldn't break anything, but I do kinda like having methods that accept domain objects instead of accepting the relevant info and creating domain objects internally

poprox-db/migrations/versions/2024_09_04_1125-8bf414e0ddfb_remove_name_column_from_datasets_table.py
karlhigley: Each experiment is owned by a team and corresponds to a dataset, which (among other things) is how we know which account aliases to use when exporting experiment data. Loading a manifest is currently the only way to create a dataset, and the manifest doesn't have a field for naming the dataset so it's not clear what to put in this column.

An alternative change would be to make this column nullable if we expect other ways of creating datasets that would allow us to provide reasonable names.

src/poprox_storage/concepts/manifest.py
karlhigley: > Is the implication here that the manifest will contain a list of member UUIDs?
Yes!

> Do we like this, or should we have the manifest list emails and then look up the UUIDs as part of ingress here?
No! I do not like this but it allowed me to punt on integrating account lookups, which complicate the picture as far as testing goes.

> would this be list[str] or list[UUID] at this point? I feel like it would be list[str] but the Team class has list[UUID]?
It should be `list[UUID]` but it's actually `list[str]`. I'll fix this!

src/poprox_storage/repositories/accounts.py
karlhigley: It should probably upsert

tests/concepts/test_manifest.py
karlhigley: Basically, yes! `project_root()` returns a `pathlib.Path`, which provides this syntax and applies the OS-appropriate magic you inferred.

poprox-db/migrations/versions/2024_09_04_1125-8bf414e0ddfb_remove_name_column_from_datasets_table.py
kluver: I think we're running into an overload of the term dataset. My gut says that outside experiment-team-tied datasets we may one-day have "public" datasets.

I don't think that hypothetical is worth keeping a column you're unsure about -- if we hit the hypothetical future where we have "public" datasets, let's solve that problem later.

src/poprox_storage/concepts/manifest.py
kluver: > The conversion from string to UUID happens when Pydantic turns the manifest JSON into model objects.

Neat!

poprox-db/migrations/versions/2024_09_04_1125-8bf414e0ddfb_remove_name_column_from_datasets_table.py
karlhigley: Public vs private is an interesting wrinkle too, but I was actually thinking of something a little different:

We've designed the database schema so that every experiment has a dataset, but not all datasets have experiments. That leaves us room to do things like export data for an experimenter so they can check properties of the data in order to find out if our platform makes sense for their experiment (like @sophiasun0515 has wanted to do re: domestic/international news bias.) Exporting data that way should also create a dataset with associated account id aliases but won't involve importing a manifest since it's pre-experiment.

PR #45, opened 2024-10-01
Shahan/onboarding/survey

ShahanM:
I created a Demographics table and a respective repo object as a DAO.

### DIFF
--- a/poprox-db/migrations/versions/2024_09_30_1755-8418372f011f_create_demographics_table.py
+++ b/poprox-db/migrations/versions/2024_09_30_1755-8418372f011f_create_demographics_table.py
@@ -7,9 +7,8 @@ Create Date: 2024-09-30 17:55:31.059281
 """
 from typing import Sequence, Union
 
-from alembic import op
 import sqlalchemy as sa
-
+from alembic import op
 
 # revision identifiers, used by Alembic.
 revision: str = '8418372f011f'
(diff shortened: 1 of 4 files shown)


karlhigley (approved):
Looks good once the linter issues are addressed

poprox-db/migrations/versions/2024_09_30_1755-8418372f011f_create_demographics_table.py
```diff
@@ -0,0 +1,52 @@
...
+    op.create_table(
+        "demographics",
+        sa.Column(
+            "demographic_id",
+            sa.UUID,
+            primary_key=True,
+            server_default=sa.text("gen_random_uuid()"),
+        ),
+        sa.Column("account_id", sa.UUID, nullable=False),
+        sa.Column("gender", sa.String(50), nullable=False),
```
kluver, line 30: I don't really think we're using fixed-length strings anywhere else in the codebase. At the same time I don't _mind_ having them, especially where we have good reason to think there is a valid upper-limit.

ShahanM: Do you want me to roll that back? I didn't put much thought into it other than the fact that it felt right - not very sound reasoning.

kluver: Nah, I think this is fine. since we have a fixed vocabulary for these.

PR #48, opened 2024-10-04
Query and export the relevant data for experiment dataset exports

karlhigley:
This is preparation for a new Lambda that exports articles, newsletters, and clicks for a single experiment. Along the way, it does a few things that make that easier:

* Creates a combined `inject_repos` decorator that can inject either database or S3 repositories
* Moves existing code for flattening domain models and writing them to Parquet files on S3 into S3 repository classes
* Adds some database repository class methods for fetching the appropriate articles, newsletters, and clicks

TODOS:

- [x] Sort out `pandas` import issue surfaced by the tests
- [x] Update the imports in the `poprox_storage.repositories` init file

### DIFF
--- a/src/poprox_storage/repositories/experiments.py
+++ b/src/poprox_storage/repositories/experiments.py
@@ -67,6 +67,23 @@ class DbExperimentRepository(DatabaseRepository):
 
         return experiment_id
 
+    def fetch_experiment_by_id(self, experiment_id: str) -> Experiment | None:
+        expt_table = self.tables["experiments"]
+        expt_query = select(expt_table).where(expt_table.c.image_id == experiment_id)
+
+        result = self.conn.execute(expt_query).first()
+        if not result:
+            return None
+        else:
+            return Experiment(
+                experiment_id=result.experiment_id,
+                owner=None,
+                description=result.description,
+                start_date=result.start_date,
+                end_date=result.end_date,
+                phases=[],
+            )
+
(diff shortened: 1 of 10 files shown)


src/poprox_storage/repositories/experiments.py
```diff
@@ -67,6 +67,22 @@ def store_experiment(
...
+        result = self.conn.execute(expt_query).first()
+        if not result:
+            return None
+        else:
+            return Experiment(
+                experiment_id=result.experiment_id,
+                description=result.description,
+                start_date=result.start_date,
+                end_date=result.end_date,
+                phases=[],
```
kluver, line 83: I'm torn on this. My gut says "if we're going to get the experiment we should get all of the experiment", but my other instinct is "if we don't need the extra data we shouldn't fetch it".

Thoughts?

kluver (approved):
I didn't see anything worth blocking on.

src/poprox_storage/repositories/experiments.py
karlhigley: I generally agree but didn't want to get hung up on fetching things that aren't used by the export since the export is the linchpin of a bunch of experiment-related work we'd also like to start on. I think there's work to do in the repository classes to make them consistently return domain concept model objects too.

kluver: Problem for another PR then.

PR #53, opened 2024-10-24
Unsubscribe: Tables altered and new functions added

aisa6148:
(Need to test after merging main)
Unsubscribe from emails
1. End consent/subscription - end date of all versions of consent to be set to NOW, remove from subscription table
2. Withdraw email from POPROX - set email field in accounts table to be NULL
3. Withdraw all data - add a column to accounts table (ifDeleted) and set it to true, UI to show "your response has been recorded", set up a weekly/monthly cleanup job to removing all behavioral data

### DIFF
--- a/src/poprox_storage/repositories/accounts.py
+++ b/src/poprox_storage/repositories/accounts.py
@@ -3,13 +3,7 @@ from datetime import date
 from uuid import UUID
 
 import sqlalchemy
-from sqlalchemy import (
-    Connection,
-    and_,
-    null,
-    or_,
-    select,
-)
+from sqlalchemy import Connection, and_, null, or_, select
 
 from poprox_concepts import Account
 from poprox_concepts.api.tracking import LoginLinkData
@@ -218,3 +212,38 @@ class DbAccountRepository(DatabaseRepository):
             data=link_data.data,
         )
         self.conn.execute(query)
+
+    def end_consent_for_account(self, account_id: UUID):
+        consent_tbl = self.tables["account_consent_log"]
+        update_query = (
+            consent_tbl.update()
+            .where(
+                consent_tbl.c.account_id == account_id,
+                consent_tbl.c.ended == null(),
+            )
+            .values(ended=sqlalchemy.text("NOW()"))
+        )
+        self.conn.execute(update_query)
+
+    def remove_email_for_account(self, account_id: UUID):
+        account_tbl = self.tables["accounts"]
+        update_query = (
+            account_tbl.update()
+            .where(
+                account_tbl.c.account_id == account_id,
+                account_tbl.c.email != null(),
+            )
+            .values(ended=null())
+        )
+        self.conn.execute(update_query)
+
+    def set_deletion_for_account(self, account_id: UUID):
+        account_tbl = self.tables["accounts"]
+        update_query = (
+            account_tbl.update()
+            .where(
+                account_tbl.c.account_id == account_id,
+            )
+            .values(ended=True)
+        )
+        self.conn.execute(update_query)


dev/dump.sql
```diff
@@ -67,10 +68,11 @@ CREATE VIEW public.account_current_interest_view AS
 
 CREATE TABLE public.accounts (
     account_id uuid DEFAULT gen_random_uuid() NOT NULL,
-    email character varying NOT NULL,
+    email character varying,
     created_at timestamp without time zone DEFAULT now() NOT NULL,
     status character varying NOT NULL,
-    source character varying NOT NULL
+    source character varying NOT NULL,
+    is_deleted boolean DEFAULT false NOT NULL
```
karlhigley, line 75: These look like reasonable changes to the database schema but they'll need to be applied in a different place. We use `alembic` to manage changes to the schema as a series of migrations—here's [an example](https://github.com/CCRI-POPROX/poprox-storage/blob/main/poprox-db/migrations/versions/2024_09_11_1524-f7953f878964_add_treatment_id_to_newsletters_table.py) of one that adds a column to an existing table. You can create a new migration from the `poprox-db` directory with `alembic revision -m "a description of the change"`

karlhigley: (This file is exported from our database as a way to provide some data that can be loaded for testing, but I think it may be out of date and/or missing some tables.)

src/poprox_storage/repositories/accounts.py
```diff
@@ -218,3 +212,38 @@ def store_login(self, link_data: LoginLinkData):
...
+
+    def end_consent_for_account(self, account_id: UUID):
+        consent_tbl = self.tables["account_consent_log"]
+        update_query = (
+            consent_tbl.update()
+            .where(
+                consent_tbl.c.account_id == account_id,
+                consent_tbl.c.ended == null(),
+            )
+            .values(ended=sqlalchemy.text("NOW()"))
```
karlhigley, line 224: I think this column is called `ended_at` above (which is a good name for it)

src/poprox_storage/repositories/accounts.py
```diff
@@ -218,3 +212,38 @@ def store_login(self, link_data: LoginLinkData):
...
+
+    def remove_email_for_account(self, account_id: UUID):
+        account_tbl = self.tables["accounts"]
+        update_query = (
+            account_tbl.update()
+            .where(
+                account_tbl.c.account_id == account_id,
+                account_tbl.c.email != null(),
+            )
+            .values(ended=null())
```
karlhigley, line 236:
```suggestion
            .values(email=null())
```

src/poprox_storage/repositories/accounts.py
```diff
@@ -218,3 +212,38 @@ def store_login(self, link_data: LoginLinkData):
...
+        self.conn.execute(update_query)
+
+    def set_deletion_for_account(self, account_id: UUID):
+        account_tbl = self.tables["accounts"]
+        update_query = (
+            account_tbl.update()
+            .where(
+                account_tbl.c.account_id == account_id,
+            )
+            .values(ended=True)
```
karlhigley, line 247:
```suggestion
            .values(is_deleted=True)
```

karlhigley (commented):
You're headed in the right direction! Left some small suggestions, but the main thing to work on is creating a database migration for your schema changes

dev/dump.sql
kluver: (to be clear, it is out of date, but we also run alembic heads on top of this to get the tables set up correctly on init so things should largely be OK regardless...)

kluver (changes requested):
Agree with karl here. If you need help with Alembic just let one of us know -- it's not hard to do by example, but finding the examples and testing can be a bear sometimes.

poprox-db/migrations/versions/2024_10_29_1406-585c9e8b4487_add_id_deleted_column_to_accounts.py
```diff
@@ -0,0 +1,26 @@
...
+
+# revision identifiers, used by Alembic.
+revision: str = "585c9e8b4487"
+down_revision: Union[str, None] = "f8667df4b049"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.add_column("accounts", sa.Column("is_deleted", sa.Boolean, nullable=True))
```
kluver, line 22: Do we want nullable here? I recognize that you need to make it nullable to add it, but do we want the "final form" of this column to be nullable, or do we want it default-false? What's the difference between `is_deleted = null` and `is_deleted=False`?

(To do this if memory serves you do an add column that's nullable, update all rows to false, then modify to be non-null)

poprox-db/migrations/versions/2024_10_29_1406-f8667df4b049_merge_head_revisions.py
```diff
@@ -0,0 +1,26 @@
+"""merge head revisions
+
+Revision ID: f8667df4b049
+Revises: 0f5739c4d79d, 9bd2ccb216a6
+Create Date: 2024-10-29 14:06:37.323505
+
+"""
+from typing import Sequence, Union
+
+from alembic import op
```
kluver, line 10: we can probably get this fixed when we want to merge. Manual fixing these isn't too hard, but I think `pre-commit run --all-files` or something similar should also fix the problem.

kluver (approved):
ltm, but a suggestion for improvement is included.

poprox-db/migrations/versions/2024_10_29_1406-585c9e8b4487_add_id_deleted_column_to_accounts.py
aisa6148: I had the same approach as well, but went the other way. I shall make the changes

karlhigley: That's not the wrong thing to do, but I suggested skipping it because of a defaults-related hassle:

SQLAlchemy treats non-nullable fields with database-level defaults as required, even though omitting them should theoretically be allowed. If this PR makes the new field non-nullable, it will also need to update the repository code to make sure that field is set whenever we store accounts.

aisa6148:
@karlhigley All checks have passed for this one (although the error sent over slack persists). Apart from that this branch is updated with main and is ready to merge!

PR #58, opened 2024-11-04
Create tables and methods for the rotating survey calendar

karlhigley:


poprox-db/migrations/versions/2024_11_04_1040-ac1a317899d9_add_survey_code_column_to_qualtrics_.py
```diff
@@ -0,0 +1,26 @@
...
+
+# revision identifiers, used by Alembic.
+revision: str = "ac1a317899d9"
+down_revision: Union[str, None] = "8f10a2101cb8"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.add_column("qualtrics_surveys", sa.Column("survey_code", sa.String(length=3), nullable=True))
```
karlhigley, line 22: I think we may as well use the existing survey code field from the spreadsheet, so I added it here and we can manually insert the codes in the existing survey rows

src/poprox_storage/repositories/qualtrics_survey.py
```diff
@@ -96,3 +100,45 @@ def store_survey_response(self, response: QualtricsSurveyResponse) -> UUID | Non
             },
             constraint="uq_qualtrics_response_id",
         )
+
```
karlhigley, line 103: I don't really like the return types on either of the following methods, but they were the best I could think of without making larger modifications to the existing domain concept classes. It seems like the ideal setup from the querying side would be for `QualtricsSurveyResponse` to have a `QualtricsSurveyInstance` as an attribute, but that may require some changes to the storing-side code.

src/poprox_storage/repositories/qualtrics_survey.py
```diff
@@ -96,3 +100,45 @@ def store_survey_response(self, response: QualtricsSurveyResponse) -> UUID | Non
...
+    def fetch_latest_survey_sent(self, date: datetime.date | None = None) -> UUID:
+        survey_calendar_table = self.tables["qualtrics_survey_calendar"]
+
+        date = date or datetime.date.today()
+
+        query = select(survey_calendar_table).where(survey_calendar_table.c.created_at <= date).limit(1)
+        row = self.conn.execute(query).fetchone()
+
+        return row.survey_id
```
karlhigley, line 144: Here I wonder if this should join with the `qualtrics_surveys` table and return a `QualtricsSurvey` object

poprox-db/migrations/versions/2024_11_04_1040-ac1a317899d9_add_survey_code_column_to_qualtrics_.py
kluver: Should we drop the length=3 limit? I don't know if that does us much good in practice since I think Shahan mentioned already running short on 3-letter-codes that are useful.

src/poprox_storage/repositories/qualtrics_survey.py
kluver: We should capture this possible refactor somewhere. I agree this feels screwy in a "model objects are at the wrong level" type of way. but I also don't know if this is the PR for that refactor :/

src/poprox_storage/repositories/qualtrics_survey.py
kluver: I think your instincts are right here -- returning UUID feels weird, despite the full survey object not being super useful for this task.

kluver (approved):
a few thoughts for changes that might be worth doing, but ultimately LGTM

src/poprox_storage/repositories/qualtrics_survey.py
```diff
@@ -96,3 +100,45 @@ def store_survey_response(self, response: QualtricsSurveyResponse) -> UUID | Non
...
+            )
+            for row in results
+        ]
+
+    def fetch_latest_survey_sent(self, date: datetime.date = None) -> UUID:
+        survey_calendar_table = self.tables["qualtrics_survey_calendar"]
+
+        date = date or datetime.date.today()
+
+        query = select(survey_calendar_table).where(survey_calendar_table.c.created_at <= date).limit(1)
```
kluver, line 141: Am I missing code to store a survey into the calendar table?

karlhigley: No, I am!

PR #59, opened 2024-11-05
Store and fetch demographic info

karlhigley:


src/poprox_storage/repositories/demographics.py
```diff
@@ -30,9 +30,10 @@ def store_demographics(self, demographic: Demographics) -> None:
                 account_id=demographic.account_id,
                 gender=demographic.gender,
                 birth_year=demographic.birth_year,
-                zip5=demographic.zip5,
+                zip3=demographic.zip5,
```
karlhigley, line 33: I think this will work but we should also change the name of the field to `zip3` here

poprox-db/migrations/env.py
```diff
@@ -27,9 +27,9 @@
 # ... etc.
 
 db_user = os.environ.get("POPROX_DB_USER", "postgres")
-db_password = os.environ.get("POPROX_DB_PASSWORD", "")
+db_password = os.environ.get("POPROX_DB_PASSWORD", "thisisapoproxpassword")
```
karlhigley, line 30: Not a big deal since these are just dev database credentials, but let's leave this out:
```suggestion
db_password = os.environ.get("POPROX_DB_PASSWORD", "")
```

poprox-db/migrations/env.py
```diff
@@ -27,9 +27,9 @@
...
 db_host = os.environ.get("POPROX_DB_HOST", "127.0.0.1")
-db_port = os.environ.get("POPROX_DB_PORT", 5432)
+db_port = os.environ.get("POPROX_DB_PORT", 5433)
```
karlhigley, line 32:
```suggestion
db_port = os.environ.get("POPROX_DB_PORT", 5432)
```

poprox-db/migrations/env.py
kluver: we should check that these changes don't break the tests. my _gut_ says that the tests don't currently change the terminal environment to plug-in the test-database (which I think is technically different from the dev-db, but is basically configured the same)

src/poprox_storage/repositories/demographics.py
```diff
@@ -56,6 +57,31 @@ def fetch_demographics_by_account_ids(self, account_ids: list[UUID]) -> list[Dem
...
+        result = self.conn.execute(demo_query).fetchone()
+
+        if result is None:
+            return None
+
+        return Demographics(
+            account_id=result.account_id,
+            gender=result.gender,
+            birth_year=result.birth_year,
+            zip5=result.zip3,
```
kluver, line 80: this line looks funky.

src/poprox_storage/repositories/accounts.py
```diff
@@ -106,6 +107,24 @@ def store_new_account(self, email: str, source: str, subsource: str = None) -> A
             account_id=row.account_id, email=row.email, status=row.status, source=source, subsource=subsource
         )
 
+
+######################### storing & fetching the zip code #########################
+    
+    def store_zip_code(self, account_id: UUID, zip5: str) -> bool:
```
kluver, line 113: given that we're using specifically zip5 vs. zip3 (and zip+4 logically does exist) should we update this to `store_zip5`?

or similar?

~Kluver

src/poprox_storage/repositories/demographics.py
kluver: Do we need an explicit truncation here?

karlhigley: It's confusing because of the attribute name (which needs to change) but `Demographics` only has `zip3` so both sides of the `=` are already `zip3` here

src/poprox_storage/repositories/demographics.py
karlhigley: Same issue here: the code is right but the attribute name is wrong

src/poprox_storage/repositories/accounts.py
Sris-Ty: Changed the naming.

PR #63, opened 2024-12-11
revert changes to dump.sql

kluver:


kluver:
I'm curious if it's maybe worth throwing a "DO NOT EDIT DIRECTLY" comment at the top to avoid this regression in the future.

karlhigley (approved):

PR #64, opened 2024-12-16
add functions to allow getting the most recent newsletter for a given survey

kluver:
- Added timestamp to the model object for a survey
- made a method to get survey instance given survey instance id
- made a method to fetch the most recent survey for a given user from a given timestamp.
  - Concern here -- I didn't bother fetching the impressions since we don't really need them for the current version of this task. By not fetching them this function stands our since every other function does fetch them. I'm curious what other people think about this.

### DIFF
--- a/src/poprox_storage/concepts/qualtrics_survey.py
+++ b/src/poprox_storage/concepts/qualtrics_survey.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from datetime import datetime
 from uuid import UUID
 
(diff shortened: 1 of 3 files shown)


karlhigley (approved):

karlhigley:
Since it's not obvious from the code that this method is for a single purpose (right now anyway), I could see not getting the impressions being potentially confusing somewhere down the line, but I don't think it's a blocker for this PR.

PR #65, opened 2024-12-16
Match only the latest version of articles by source and external id

karlhigley:
It seems like this join is pulling back too many articles with the same source (`AP`) and external id, which appears to be causing some of our issues with duplicate articles ending up in people's recommendations.


kluver (approved):
lgtm

PR #67, opened 2024-12-18
Handling multiple images in articles and images

Mahamudul42:
Handling multiple images in articles and images

### DIFF
--- a/poprox-db/migrations/versions/2024_12_11_1611-c2007fa35fc2_add_caption_column_to_images_table.py
+++ b/poprox-db/migrations/versions/2024_12_11_1611-c2007fa35fc2_add_caption_column_to_images_table.py
@@ -7,9 +7,8 @@ Create Date: 2024-12-11 16:11:30.234857
 """
 from typing import Sequence, Union
 
-from alembic import op
 import sqlalchemy as sa
-
+from alembic import op
 
 # revision identifiers, used by Alembic.
 revision: str = 'c2007fa35fc2'
--- a/poprox-db/migrations/versions/2024_12_11_1611-c2007fa35fc2_add_caption_column_to_images_table.py
+++ b/poprox-db/migrations/versions/2024_12_11_1611-c2007fa35fc2_add_caption_column_to_images_table.py
@@ -1,0 +1,26 @@
+"""add caption column to images table
+
+Revision ID: c2007fa35fc2
+Revises: 03883e546c24
+Create Date: 2024-12-11 16:11:30.234857
+
+"""
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision: str = 'c2007fa35fc2'
+down_revision: Union[str, None] = '03883e546c24'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.add_column("images", sa.Column("caption", sa.String, nullable=True))
+
+
+def downgrade() -> None:
+    op.drop_column("images", "caption")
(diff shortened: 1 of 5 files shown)


poprox-db/migrations/versions/2024_12_11_1611-c2007fa35fc2_add_caption_column_to_images_table.py
```diff
@@ -0,0 +1,26 @@
+"""add caption column to images table
+
```
karlhigley, line 2: It doesn't seem like you have the pre-commit hooks installed, which would raise these issues when you try to commit. You can install them by running `pre-commit install` in each repo directory.

src/poprox_storage/repositories/articles.py
```diff
@@ -31,6 +32,7 @@ def __init__(self, connection: Connection):
             "articles",
             "entities",
             "mentions",
+            "article_image_associations",
```
karlhigley, line 35: Not sure about the name of this table, but also not sure we have another join table like this that has established a convention.

poprox-db/migrations/versions/2024_12_11_1633-adf30c8ce87e_create_image_associations_table.py
```diff
@@ -0,0 +1,49 @@
...
+        ["article_id"],
+    )
+
+    op.create_foreign_key(
+        "fk_image_associations_image_id",
+        "article_image_associations",
+        "images",
+        ["image_id"],
+        ["image_id"],
+    )
```
karlhigley, line 43: This table should also have a uniqueness constraint to prevent duplicate rows with the same image and article ids

karlhigley (commented):
Looks good, aside from a missing constraint and some formatting issues

poprox-db/migrations/versions/2024_12_11_1611-c2007fa35fc2_add_caption_column_to_images_table.py
Mahamudul42: pre-commit installed now.

poprox-db/migrations/versions/2024_12_11_1633-adf30c8ce87e_create_image_associations_table.py
Mahamudul42: The following constraint has been added for your review.

op.create_unique_constraint(
        "uq_article_image_associations",
        "article_image_associations",
        ["article_id", "image_id"]
    )

poprox-db/migrations/versions/2024_12_11_1633-adf30c8ce87e_create_image_associations_table.py
```diff
@@ -0,0 +1,54 @@
...
+    op.create_unique_constraint(
+        "uq_article_image_associations",
+        "article_image_associations",
+        ["article_id", "image_id"]
```
karlhigley, line 46: That looks like the right set of arguments, but there's one minor syntax thing. I think the last one should be a `tuple` with `()` instead of a `list` with `[]`:
```suggestion
        ("article_id", "image_id")
```

Mahamudul42: Changed with your suggestions. Thank you.

poprox-db/migrations/versions/2024_12_11_1611-c2007fa35fc2_add_caption_column_to_images_table.py
karlhigley: It should be flagging the pre-commit issues when you try to commit, but it doesn't seem like that's working. To run the pre-commit formatters and checks manually, you can do `pre-commit run --all-files` in the repo directory

karlhigley (approved):

PR #71, opened 2025-01-30
Placebo `rec_id`.

sushobhan2024:
This PR–
_Updated:_
1. Adds DB migration for `placebo_id`. 
2. Sets `placebo_id` and stores it in the accounts.
3. Updates `placebo_id` with a new one when a user opts out of experiment.


src/poprox_storage/repositories/accounts.py
```diff
@@ -282,9 +282,9 @@ def set_deletion_for_account(self, account_id: UUID):
         )
         self.conn.execute(update_query)
 
-    def set_placebo_rec_id(self, account_id: UUID):
-        rec_id = f"{str(date.today().year)[-2:]}.{str(date.today().month).zfill(2)}.{account_id.hex[:6]}"
+    def set_placebo_id(self, account_id: UUID):
+        placebo_id = f"{str(date.today().year)[-2:]}.{str(date.today().month).zfill(2)}.{account_id.hex[:6]}"
```
karlhigley, line 286: The group ids we use when subscribers are in experiments are UUIDs, so either those should change to match this format or we should use a UUID here. Or maybe both?

I do kind of like the readability of this format, but this method seems like it's handling presentation logic in the database code. What if we made the `placebo_id` column a UUID, and handled formatting it this way when we render the email template?

kluver: I'm in a similar boat to Karl here, but I think I have conflicting recommendations about next step. I like the readability of this format, which makes me think this is not a suitable `placebo_id` unless. Unless I'm mistaken, the placebo is meant to be just plain fake/random data. This doesn't look random and I worry people will pretty quickly understand the year/month part of it and notice the last bit isn't changing.

We could make this a UUID per karl's suggestion or we could make this literally 5 random letters and numbers. If we want to keep the year/month bit we could maybe do that in the email template per karl's suggestion.

I almost feel like "liking this format" is a sign that we haven't implemented what Joe put in the IRB documentation.

kluver (changes requested):

src/poprox_storage/repositories/accounts.py
karlhigley: I think we can punt on the question of what a proper format is to the platform PR if the `placebo_id` column becomes a UUID to make it interchangeable with experiment group ids. Then we can apply the year/month format or some conversion of a UUID into a short string downstream.

kluver: So long as we recognize the format should be more random and less transparent I'm happy.

By punt do you mean "let's get the other PR sorted and then return to this" or do you mean "merge this and return later?"

karlhigley: Once `placebo_id` becomes a UUID, then I think we can merge this. We can resolve the question of whether to display a UUID or something else when we get to `platform`'s presentation layer in the other PR.

sushobhan2024: @karlhigley @kluver Thank you for the comments! I have now updated `placebo_id` to be a UUID, directly using `uuid4()` in the code. I was looking at the code base to see where and how `group_id`s are generated so I can use similar way to generate UUIDs for `placebo_id`s. `experiment.py` fetches `group_id` while `poprox_storage.concepts.experiment.py` is validating, right?

src/poprox_storage/repositories/experiments.py
```diff
@@ -230,6 +231,10 @@ def update_expt_assignment_to_opt_out(self, account_id: UUID) -> UUID | None:
         )
         self.conn.execute(assignment_query)
 
+        # update placebo id
+        account_repo = DbAccountRepository()
+        account_repo.set_placebo_id(account_id)
+
```
karlhigley, line 237: I've been trying to avoid instantiating repository classes within the code of other repositories, so I'd probably move this behavior out to wherever `update_expt_assignment_to_opt_out()` gets called (which is somewhere in `platform` or `web`, but I forget which.)

sushobhan2024: I will make this change. It's in `app.py` in `web`.

karlhigley (approved):

kluver (approved):
From what I can tell this meets the requirements. I'm happy

PR #72, opened 2025-01-31
Add experiment extra column to impressions table and update fetch

sophiasun0515:
For the new column adding, it passed local db test.
This is associated with the data model change in https://github.com/CCRI-POPROX/poprox-concepts/pull/37 (I think it's also why the first PyTest test failed since that PR is not merged yet).


karlhigley (approved):

kluver (approved):

karlhigley:
Code looks fine but let's hold off on merging this until we have the recommender end of the problem sorted out

karlhigley (changes requested):
Code looks fine but let's hold off on merging this until we have the recommender end of the problem sorted out

PR #73, opened 2025-01-31
Add preview image id to impressions table

Mahamudul42:
1. Newsletter class is modified for preview images.
2.  Migration file is added to insert a new column.


karlhigley:
It looks like one of the tests needs an update to include a preview image id when it creates an `Impression` object

kluver:
> It looks like one of the tests needs an update to include a preview image id when it creates an `Impression` object

@Mahamudul42 -- let us know if you need help getting the tests to run locally by-the-way. I can't remember if we got that working on your dev environment or not, but I remember running into trouble with that on _someone's_ computer lately.

Without the tests passing I don't think it's a good idea to merge this.

karlhigley (approved):

PR #74, opened 2025-02-06
Kluver/feature/qualtrics question metadata

kluver:
- merge split heads for alembic
- fix typo in past revision to make tests pass
- add a BSON column for qualtrics survey question metadata.

### DIFF
--- a/poprox-db/migrations/versions/2025_02_06_1339-6da1373105ef_add_column_for_qualtrics_question_.py
+++ b/poprox-db/migrations/versions/2025_02_06_1339-6da1373105ef_add_column_for_qualtrics_question_.py
@@ -23,4 +23,4 @@ def upgrade() -> None:
 
 
 def downgrade() -> None:
-    op.drop_column("qualtrics_surveys", "question_metadata")
+    op.drop_column("qualtrics_surveys", "question_metadata_raw")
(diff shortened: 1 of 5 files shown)


kluver:
Converting to draft -- I thought the survey concept was in concepts -- but it seems it's in survey so  I don't need to split this into so many small PRs.

kluver:
OK, that should be the whole change now.

poprox-db/migrations/versions/2025_02_06_1339-6da1373105ef_add_column_for_qualtrics_question_.py
```diff
@@ -0,0 +1,26 @@
...
+
+# revision identifiers, used by Alembic.
+revision: str = "6da1373105ef"
+down_revision: Union[str, None] = "a67fedcb0af0"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.add_column("qualtrics_surveys", sa.Column("question_metadata", JSONB, nullable=True))
```
kluver, line 22: I think there's a discussion to be had here around what we name this column. I think there's an argument for `question_metadata_raw` I think a lot of similar database columns have `_raw` as part of the name.

Open to change, looking for a second oppinion

karlhigley (approved):

poprox-db/migrations/versions/2025_02_06_1339-6da1373105ef_add_column_for_qualtrics_question_.py
kluver, line 22:
```suggestion
    op.add_column("qualtrics_surveys", sa.Column("question_metadata_raw", JSONB, nullable=True))
```

kluver:
I thought I would just do the chance to `question_metadata_raw` in github, but I just realized how many lines that is. new commit coming in a few minutes then I think we're ready to merge.

PR #75, opened 2025-02-10
Remove database commit management and account alias creation from experiment/dataset repos

karlhigley:
The experiment side of this used to be a stand-alone "one stop shop" for creating an experiment and everything involved in that, but we've since split out team/dataset/alias creation into separate repos. In order to avoid partially creating the relevant database records when loading an experiment manifest, the database commit management needs to happen at the Lambda level.

### DIFF
--- a/src/poprox_storage/repositories/datasets.py
+++ b/src/poprox_storage/repositories/datasets.py
@@ -14,12 +14,10 @@ class DbDatasetRepository(DatabaseRepository):
         )
 
     def store_new_dataset(self, accounts: list[Account], team_id: UUID) -> UUID:
-        self.conn.commit()
-        with self.conn.begin():
-            dataset_id = self._insert_dataset(team_id)
+        dataset_id = self._insert_dataset(team_id)
 
-            for account in accounts:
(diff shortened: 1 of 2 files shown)


kluver (approved):

kluver:
As I was reviewing I double checked the functions that deligate to `_insert_model(...)` They all looked good too, however, if we're moving to standardize commit and transaction management to the lambda level, should the `commit` parameter to `_insert_model` be made `False` default? Right now all calls in the two files you showed here explicitly set `commit=False` when calling `_insert_model`.

karlhigley:
Maybe, but it's a bigger change than I want to tackle here. In this PR, `commit` is always set to `False` so that the Lambda can abort the transaction and clean up after itself if there's an error in the middle. Ingesting an experiment manifest has to set up a complicated web of rows in different tables accessed with different repositories though, and most other places are fine with inserting a model being an atomic operation by itself (e.g. `store_article` just creates one row in one table.)

kluver:
> Maybe, but it's a bigger change than I want to tackle here. 

Super fair -- just thinking aloud

PR #76, opened 2025-03-06
change how init_ and stop_poprox_dev modify their paths to start/stop the db container

kluver:


kluver:
This test has failed. We're going to close the PR until we ind a different way to do this

PR #79, opened 2025-03-13
added method to fetch assignments by experiment_id

ShahanM:
Contains 
 * a method to fetch assignments by experiment id, and refactor of the existing `fetch_active_expt_assignments` to reduce code duplication.
* Code to fetch the full object graph of an experiment (and small test updates related to this work)
* a function to store a new qualtrics surveys

### DIFF
--- a/src/poprox_storage/repositories/experiments.py
+++ b/src/poprox_storage/repositories/experiments.py
@@ -182,10 +182,21 @@ class DbExperimentRepository(DatabaseRepository):
         return recommender_lookup_by_group
 
     def fetch_active_expt_assignments(self, date: datetime.date | None = None) -> dict[UUID, Assignment]:
-        assignments_tbl = self.tables["expt_assignments"]
-
         group_ids = self.fetch_active_expt_group_ids(date)
+        group_lookup_by_account = self._fetch_assignments_by_group_ids(group_ids)
 
+        return group_lookup_by_account
+
+    def fetch_assignments_by_experiment_id(self, experiment_id: UUID) -> dict[UUID, Assignment]:
+        experiment = self.fetch_experiment_by_id(experiment_id)
+
+        group_ids = [group.id for group in experiment.groups]
+        group_lookup_by_account = self._fetch_assignments_by_group_ids(group_ids)
+
+        return group_lookup_by_account
+
+    def _fetch_assignments_by_group_ids(self, group_ids: list[UUID]) -> dict[UUID, Assignment]:
+        assignments_tbl = self.tables["expt_assignments"]
         group_query = select(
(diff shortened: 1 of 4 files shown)


karlhigley (approved):

src/poprox_storage/repositories/experiments.py
```diff
@@ -182,10 +182,21 @@ def fetch_active_expt_recommender_urls(self, date: datetime.date | None = None)
...
-
         group_ids = self.fetch_active_expt_group_ids(date)
+        group_lookup_by_account = self._fetch_assignments_by_group_ids(group_ids)
+
+        return group_lookup_by_account
+
+    def fetch_assignments_by_experiment_id(self, experiment_id: UUID) -> dict[UUID, Assignment]:
+        experiment = self.fetch_experiment_by_id(experiment_id)
 
+        group_ids = [group.id for group in experiment.groups]
```
kluver, line 193: I don't think fetch_experiment_by_id returns groups.

kluver: (sorry for taking so long to notice this issue. I was doubling back to review PRs before the Tuesday meeting.)
I might want @karlhigley's double-check here, but the implementation of `fetch_experiment_by_id` that  currently see sets the phases to `[]` an I think `.groups` pulls the group info out of the phase dictionary right? I'm not missing somewhere where this info is filled in later right?

karlhigley: I don't think we've had any reason to fetch an experiment with the whole object graph pulled along with it before, so yeah, this won't currently ever return the groups. We could modify the experiment fetching method to pull in the phases etc though

kluver: I think that would be my current proposal for how to make this work. I know that' a few extra queries, but in the cosmic scale, I think doing that work every time is better than having lots of small functions for what parts of the object graph we want and having to remember if an experiment view model is complete or not.

kluver:
I'm a bit lost on the status of this. I think the issue is the downstream issue around pulling the whole model graph of the experiment? @ShahanM is that right?

kluver:
This PR is getting over-full. That said, since all I've added is code that should address a major weakness in the original draft and I think it should let it get commited.

I'm going to re-request @karlhigley -- I think you're the only one other than _maybe_ Michael who can review my changes with context. They are tested so I'm not super-worried.

kluver:
also RE: failing tests. I think this is actually an issue with poprox_concepts. or at least I can say with confidence there's an error there -- until we resolve that we can't get the full tests runes to know if a quality-of-life change I made broke anything else.

karlhigley (approved):

PR #82, opened 2025-03-20
Add a function `fetch_survey_metadata` in `DbQualtricsSurveyRepository`.

sushobhan2024:
This function fetches survey metadata for  the given list of `survey_ids` from the `qualtrics_surveys` table and returns a list of `QualtricsSurvey` objects.

### DIFF
--- a/src/poprox_storage/repositories/qualtrics_survey.py
+++ b/src/poprox_storage/repositories/qualtrics_survey.py
@@ -286,6 +286,31 @@ class DbQualtricsSurveyRepository(DatabaseRepository):
             question_metadata_raw=row.question_metadata_raw,
         )
 
+    def fetch_survey_metadata(self, survey_ids: list[UUID]) -> list[QualtricsSurvey]:
+        survey_table = self.tables["qualtrics_surveys"]
+
+        query = select(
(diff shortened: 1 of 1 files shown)


karlhigley (approved):

kluver:
Ope -- I didn't see this review before commenting in the other PR.

Is this function really named correctly? it seems like it's just `fetch_surveys`?

sushobhan2024:
@kluver This function returns surveys for a list of `survey_ids`. Main purpose of this was to extract `question_metadata_raw`. I can see that `fetch_surveys` does that too. I am thinking now this could have been done with `fetch_surveys`.

kluver:
I don't think I was clear properly:

1. I understand that this function had an _intent_
2. I'm worried about _code duplication_ here.
3. I'm worried about function _names_ -- this code doesn't seem to _actually_ fetch just the metadata, so why is it named that? What makes this different from a more generic function that fetches multiple surveys by ID? If the answer is "nothing" then maybe we should rename it to the more neutral `fetch_surveys` (which is not currently a function -- we have `fetch_survey` for one survey, but no pre-built method for fetching multiple surveys)

PR #85, opened 2025-03-25
Fix extra flatten by collecting all possible keys first

sophiasun0515:


src/poprox_storage/repositories/data_stores/s3.py
```diff
@@ -59,7 +59,13 @@ def _write_records_as_parquet(
...
         file_name = f"{file_prefix}_{start_time.strftime('%Y%m%d-%H%M%S')}.parquet"
 
-        arrow_table = pa.Table.from_pylist(records)
+        all_keys = set()
+        for record in records:
+            all_keys.update(record.keys())
+
+        # Define a schema that includes all possible fields
+        schema = pa.schema([pa.field(key, pa.string()) for key in all_keys])
+        arrow_table = pa.Table.from_pylist(records, schema=schema)
```
karlhigley, line 68: I think this will need to collect the keys and types in the loop because not all of our fields are strings. That probably means that `all_keys` should become `all_fields` and the default should be an empty dictionary, so that we can build up key/value pairs that have the field name and the field type.

PR #88, opened 2025-04-03
Store `pipeline_name` to `newsletter` table.

sushobhan2024:


karlhigley:
These PRs look like they modify newsletters and impressions somewhat inconsistently. I think information about what recommender was used to generate a newsletter should probably go in the newsletter table and on the `Newsletter` domain model.

PR #89, opened 2025-04-10
updated tables and added new functions

Sris-Ty:
updated newsletter and impression tables and add new function for storing feedback data

### DIFF
--- a/src/poprox_storage/repositories/newsletters.py
+++ b/src/poprox_storage/repositories/newsletters.py
@@ -9,6 +9,7 @@ from sqlalchemy import (
     insert,
     null,
     select,
+    update,
 )
 
 from poprox_concepts.domain import Account, Article, Impression, Newsletter
@@ -55,6 +56,41 @@ class DbNewsletterRepository(DatabaseRepository):
                     extra=impression.extra,
                 )
                 self.conn.execute(stmt)
+
+    def store_newsletter_feedback(self, account_id: list[UUID], newsletter_id: list[UUID], ispositive: bool):
+        newsletter_table = self.tables["newsletters"]
+
+        self.conn.commit()
+        with self.conn.begin():
+            stmt = (
+                update(newsletter_table)
+                .where(
+                    and_(newsletter_table.c.newsletter_id == newsletter_id, newsletter_table.c.account_id == account_id)
+                )
+                .values(
+                    feedback=ispositive,
+                )
+            )
+            self.conn.execute(stmt)
+
+    def store_impression_feedback(self, article_id: list[UUID], newsletter_id: list[UUID], ispositive: bool):
+        impressions_table = self.tables["impressions"]
+
+        self.conn.commit()
+        with self.conn.begin():
+            stmt = (
+                update(impressions_table)
+                .where(
+                    and_(
+                        impressions_table.c.newsletter_id == newsletter_id,
+                        impressions_table.c.article_id == article_id,
+                    )
+                )
+                .values(
+                    feedback=ispositive,
+                )
+            )
+            self.conn.execute(stmt)
 
     def fetch_newsletters(self, accounts: list[Account]) -> list[Newsletter]:
         newsletters_table = self.tables["newsletters"]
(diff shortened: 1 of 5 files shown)


src/poprox_storage/repositories/newsletters.py
```diff
@@ -56,6 +57,41 @@ def store_newsletter(self, newsletter: Newsletter):
...
 
+    def store_newsletter_feedback(self, account_id: list[UUID], newsletter_id: list[UUID], ispositive: bool):
+        newsletter_table = self.tables["newsletters"]
+
+        self.conn.commit()
+        with self.conn.begin():
+            stmt = (
+                update(newsletter_table)
+                .where(
+                    and_(newsletter_table.c.newsletter_id == newsletter_id, newsletter_table.c.account_id == account_id)
```
kluver, line 68: why use both newsletter and account id here?

src/poprox_storage/repositories/newsletters.py
```diff
@@ -56,6 +57,41 @@ def store_newsletter(self, newsletter: Newsletter):
                 )
                 self.conn.execute(stmt)
 
+    def store_newsletter_feedback(self, account_id: list[UUID], newsletter_id: list[UUID], ispositive: bool):
+        newsletter_table = self.tables["newsletters"]
+
+        self.conn.commit()
```
kluver, line 63: @karlhigley I _think_ I remember hearing that you had moved to prefer commits not occur inside repositories, am I remembering that right?

src/poprox_storage/repositories/newsletters.py
```diff
@@ -56,6 +57,41 @@ def store_newsletter(self, newsletter: Newsletter):
                 )
                 self.conn.execute(stmt)
 
+    def store_newsletter_feedback(self, account_id: list[UUID], newsletter_id: list[UUID], ispositive: bool):
```
kluver, line 60: - why lists?
- And should `ispositive` be `is_positive` for code style?
- And should it be bool|none to allow updating back to null?

src/poprox_storage/repositories/newsletters.py
```diff
@@ -56,6 +57,41 @@ def store_newsletter(self, newsletter: Newsletter):
...
+                .where(
+                    and_(newsletter_table.c.newsletter_id == newsletter_id, newsletter_table.c.account_id == account_id)
+                )
+                .values(
+                    feedback=ispositive,
+                )
+            )
+            self.conn.execute(stmt)
+
+    def store_impression_feedback(self, article_id: list[UUID], newsletter_id: list[UUID], ispositive: bool):
```
kluver, line 76: same comments on parameter types here.

kluver: Unrelated to previous,

Do we want to identify impressions by article_id and newsletter_id, or is there a dataflow where we can have them identified by impression_id directly. It's a pretty minor difference in practice, but I could see the version using impression_id directly being slightly better.

If it's not easy, I don't think it's worth changing this -- but if is easy I could see it being worth it.

src/poprox_storage/repositories/newsletters.py
```diff
@@ -103,20 +139,36 @@ def fetch_newsletters_by_treatment_id(self, expt_treatment_ids: list[UUID]) -> l
...
-            impressions_table.c.extra,
-        ).where(
-            impressions_table.c.newsletter_id.in_(newsletter_ids),
+        query = (
+            select(
+                impressions_table.c.newsletter_id,
+                impressions_table.c.article_id,
+                impressions_table.c.position,
+                impressions_table.c.extra,
+                articles_table.c.headline,
```
kluver, line 150: I'm unsure about this one -- I think the idea is that we would want to load the headline, and subhead from the impression table if it's there (it's possible that this code and the code to make it be there happened around the same time....)

I _thnk_ some legacy data might have null for headline and subhead in the impression column -- in which case we might want to back-fill from article?

src/poprox_storage/repositories/newsletters.py
Sris-Ty: I am not completely sure about that either. I used article_id and newsletter_id just to be in safe spot.

src/poprox_storage/repositories/newsletters.py
Sris-Ty: I can 't remember I added this part but I think back-fill is a good idea.

src/poprox_storage/repositories/newsletters.py
karlhigley: If you need transaction management, then yes. Here it's just one update that will be an atomic change anyway, so I don't think we need to `self.conn.begin()` in this method, which means we can also avoid the pre-emptive `commit()` that gets us outside sqlalchemy's automatic transaction management

src/poprox_storage/repositories/newsletters.py
karlhigley: Off the top of my head, the only case I can see where it would make a difference is if the same article is recommended in multiple positions—which seems unlikely to be intentional but could happen. I think I agree that switching to `impression_id` would probably be good, but if it's hard then probably okay to postpone for later.

src/poprox_storage/repositories/newsletters.py
karlhigley: I think this was written in parallel with the change that recorded headline/subhead in the impressions table, which probably should have made updates here but didn't. I'm not sure how straightforward it is to return both here since the column names will collide. If we have to pick one, probably good to use the columns from `impressions` instead of `articles`

src/poprox_storage/repositories/newsletters.py
karlhigley: Code below is correct, but the type annotations need changes. Agreed that `is_positive` would be better.

karlhigley (approved):

src/poprox_storage/repositories/newsletters.py
```diff
@@ -102,22 +136,38 @@ def fetch_newsletters_by_treatment_id(self, expt_treatment_ids: list[UUID]) -> l
...
+                impressions_table.c.newsletter_id.in_(newsletter_ids),
+            )
         )
         rows = self.conn.execute(query).fetchall()
         return [
             Impression(
                 newsletter_id=row.newsletter_id,
-                article_id=row.article_id,
+                headline=row.headline,
+                subhead=row.subhead,
```
karlhigley, line 164: We can drop the duplicate headline/subhead args here:
```suggestion
```

PR #91, opened 2025-04-10
Record recommender pipeline info.

sushobhan2024:
Add recommender pipeline name, version, and hash to `newsletters` table.


karlhigley:
The code looks fine overall but causes a circular import error. Once that's fixed we can go ahead and merge

sushobhan2024:
Yes, I noticed that the failure is because `Newsletter` model did not have `recommender_info`, which is added by the other PR.

karlhigley:
The other PR has been merged and I re-ran the tests, so this is a new failure to troubleshoot

kluver:
Manually re-running tests now that we've addressed the circular deps.

kluver:
new problem trouble-shot. I'll do a quick review while I'm here then I think we merge.

sushobhan2024:
@kluver @karlhigley Thank you very much!

PR #92, opened 2025-04-17
Added headline and subhead to newsletters.parquet export

Mahamudul42:
Added headline and subhead to newsletters.parquet export


karlhigley (approved):

karlhigley:
Thanks!

PR #93, opened 2025-04-17
Adjust subscriber query and rename some columns in data exports

karlhigley:
This is based on Ed's feedback on exported data, and includes a small fix to the determination of which users should be included in the export. Namely, anyone who _was_ subscribed during the last 30 days should be included (not just people with currently active subscriptions), so this adds a method to find those accounts.


kluver (approved):
lgtm :ship:

PR #95, opened 2025-04-28
Create candidate pool tables and create pool repository

karlhigley:
This will allow us to explicitly record what we considered to be part of the candidate pool each day rather than inferring what was part of the pool from the date. I think this will help us avoid including the same article multiple days in a row, and also make it easier to do offline evaluation down the road.

### DIFF
--- a/src/poprox_storage/repositories/articles.py
+++ b/src/poprox_storage/repositories/articles.py
@@ -42,25 +42,25 @@ class DbArticleRepository(DatabaseRepository):
         article_table = self.tables["articles"]
         cutoff = datetime.now() - timedelta(days=days_ago)
         query = select(article_table).where(article_table.c.published_at > cutoff)
-        return self._get_articles(query)
+        return _fetch_articles(self.conn, query)
 
     def fetch_articles_before(self, days_ago=1) -> list[Article]:
         article_table = self.tables["articles"]
         cutoff = datetime.now() - timedelta(days=days_ago)
         query = select(article_table).where(article_table.c.published_at < cutoff)
-        return self._get_articles(query)
+        return _fetch_articles(self.conn, query)
 
     def fetch_articles_ingested_since(self, days_ago=1) -> list[Article]:
         article_table = self.tables["articles"]
         cutoff = datetime.now() - timedelta(days=days_ago)
         query = select(article_table).where(article_table.c.created_at > cutoff)
-        return self._get_articles(query)
+        return _fetch_articles(self.conn, query)
 
     def fetch_articles_ingested_before(self, days_ago=1) -> list[Article]:
         article_table = self.tables["articles"]
         cutoff = datetime.now() - timedelta(days=days_ago)
         query = select(article_table).where(article_table.c.created_at < cutoff)
-        return self._get_articles(query)
+        return _fetch_articles(self.conn, query)
 
     def fetch_articles_ingested_between(self, start_date, end_date) -> list[Article]:
         article_table = self.tables["articles"]
@@ -71,12 +71,12 @@ class DbArticleRepository(DatabaseRepository):
                 article_table.c.created_at >= start_date,
             )
         )
-        return self._get_articles(query)
+        return _fetch_articles(self.conn, query)
 
     def fetch_articles_by_id(self, ids: list[UUID]) -> list[Article]:
         article_table = self.tables["articles"]
         query = select(article_table).where(article_table.c.article_id.in_(ids))
-        return self._get_articles(query)
+        return _fetch_articles(self.conn, query)
 
     def fetch_article_by_external_id(self, id_: str) -> Article:
         article_table = self.tables["articles"]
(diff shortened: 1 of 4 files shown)


src/poprox_storage/repositories/articles.py
```diff
@@ -228,6 +241,15 @@ def store_articles(self, articles: list[Article], *, mentions=False, progress=Fa
 
         return failed
 
+    def store_candidates(self, articles: list[Article], date: date):
+        candidates_table = self.tables["candidate_articles"]
+        insert_stmt = (
+            insert(candidates_table)
+            .values([{"article_id": article.article_id, "date": date} for article in articles])
```
kluver, line 248: I don't love `date` here to identify which newsletter batch we're talking about. But I like it better than every way I can see to fix it.

Realistically it's fine so long as we make sure never to send two newsletters in one calendar day.

src/poprox_storage/repositories/articles.py
```diff
@@ -228,6 +241,15 @@ def store_articles(self, articles: list[Article], *, mentions=False, progress=Fa
...
+            .on_conflict_do_nothing(constraint="uq_candidate_articles")
```
kluver, line 249: Was just wondering if we could have a conflict in a "newsletter crashes and we manually re-start the workflow" situation, but I think this handles that.

kluver (approved):

src/poprox_storage/repositories/articles.py
karlhigley: One thing we could do is give each candidate set its own id and then link newsletters to the specific candidate set they used

src/poprox_storage/repositories/articles.py
karlhigley: Candidate set ids would also help here (even though the uniqueness constraint mitigates the damage)

karlhigley:
Right now, we pull candidate articles that were ingested in the last 24 hours. Since we ingest candidates once per day, that will implicitly pull the articles we ingested whenever we last pulled content from the external API. I think we _could_ keep doing that, but my intent with recording the candidate set is remove ambiguity and reduce edge cases where articles make it into the pool two days in a row.

So, instead, we could fetch the most recent candidate set and know that we're using exactly the same articles we used yesterday (if it's in the morning) or the same articles we used in today's newsletter (if it's in the afternoon.) That seems a bit cleaner, and it's going to improve the downstream situation with our data exports because we will no longer have to infer which articles were eligible to be recommended based on when they were ingested.

src/poprox_storage/repositories/pools.py
```diff
@@ -0,0 +1,71 @@
...
+            {"candidate_pool_id": candidate_pool_id, "pool_type": pool_type}
+        )
+        self.conn.execute(set_insert_stmt)
+
+        insert_stmt = (
+            insert(candidate_articles_table)
+            .values(
+                [{"candidate_pool_id": candidate_pool_id, "article_id": article.article_id} for article in articles]
+            )
+            .on_conflict_do_nothing(constraint="uq_candidate_articles")
```
kluver, line 41: I don't think `on_conflict_do_nothing` should do anything here outside of error cases:

- two articles passed in. Shouldn't happen, this is as good of a plan as any.
- Duplicate uuid4. Not worth adjusting plans for.

Probably fine to keep it, or make it an error because if it happens something screwy is going on.

src/poprox_storage/repositories/pools.py
```diff
@@ -0,0 +1,71 @@
...
+
+        query = (
+            select(candidate_articles_table)
+            .join(articles_table, candidate_articles_table.c.article_id == articles_table.c.article_id)
+            .where(candidate_articles_table.c.candidate_pool_id == candidate_pool_id)
+        )
+
+        return _fetch_articles(self.conn, query)
+
+    def fetch_latest_pool_of_type(self, candidate_pool_type: str) -> UUID:
```
kluver, line 59: `type` system is a nice bit of flair here. I don't think we have a use case right now, but I could see a lot of ideas in the future that we could attach to this system without much extra architecture work.

kluver (approved):
Agreed. This is better. The old version _worked_ but this is _nice_.

src/poprox_storage/repositories/pools.py
karlhigley: Yeah, I put it here because initially thought we'd want to distinguish daily newsletter candidate pools from sign-up candidate pools. In hindsight, we can just use the latest daily newsletter candidate pool for first sign-up newsletters though. One situation in which it would make a difference would be if we switched to pulling articles from AP once an hour—in that case, we could use news from the last 24 hours for sign-up newsletters and they'd have separate candidate pools. Don't know if we want to do that, but that's the line I was thinking along.

src/poprox_storage/repositories/pools.py
karlhigley: I think the impact is just that passing the same article in twice will silently ignore the duplication. I can't see how that would happen and I'm not sure whether it's better to raise an error (or not.)

src/poprox_storage/repositories/pools.py
karlhigley: Another one could be candidate pools from multiple sources (e.g. AP vs Reuters), but that seems much farther off than pulling hourly might be and we'd probably want a different and more specific field for that

PR #101, opened 2025-05-12
Add `S3PanelManagementRepository` to store accounts for panel management.

sushobhan2024:
Add `S3PanelManagementRepository` to store accounts for panel management and update `fetch_accounts`.

### DIFF
--- a/src/poprox_storage/repositories/accounts.py
+++ b/src/poprox_storage/repositories/accounts.py
@@ -5,7 +5,7 @@ from uuid import UUID, uuid4
 import sqlalchemy
 from sqlalchemy import Connection, and_, null, or_, select
 
-from poprox_concepts import Account, AccountPanelManagement
+from poprox_concepts import Account
 from poprox_concepts.api.tracking import LoginLinkData
 from poprox_storage.repositories.data_stores.db import DatabaseRepository
 
--- a/src/poprox_storage/repositories/accounts.py
+++ b/src/poprox_storage/repositories/accounts.py
@@ -5,7 +5,7 @@ from uuid import UUID, uuid4
 import sqlalchemy
 from sqlalchemy import Connection, and_, null, or_, select
 
-from poprox_concepts import Account, AccountPanelManagement
+from poprox_concepts import Account
 from poprox_concepts.api.tracking import LoginLinkData
 from poprox_storage.repositories.data_stores.db import DatabaseRepository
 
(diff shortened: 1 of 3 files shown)


src/poprox_storage/repositories/accounts.py
```diff
@@ -51,6 +51,31 @@ def fetch_accounts(self, account_ids: list[UUID] | None = None) -> list[Account]
...
+
+        return [
+            AccountPanelManagement(
+                account_id=row.account_id,
+                source=row.source,
+                subsource=row.subsource,
+            )
+            for row in result
+        ]
+
```
karlhigley, line 78: I think the existing function that fetches accounts by id will work for what you need. The place to throw out data that shouldn't be exported is probably in the export rather than in the fetching

sushobhan2024: I am now using the existing `fetch_accounts` and filter in the export function (added in the PR in `platform`). Since I wanted to store this account export in a S3 bucket and we did not have a separate S3 bucket for panel management, I created a `S3PanelManagementRepository` in this PR. But I'm not sure if that's what we want, do we want to export it to an existing S3 bucket instead?

src/poprox_storage/repositories/accounts.py
```diff
@@ -47,6 +48,7 @@ def fetch_accounts(self, account_ids: list[UUID] | None = None) -> list[Account]
                 email=row.email,
                 status=row.status,
                 source=row.source,
+                subsource=row.subsource,
```
kluver, line 51: I think this change might conflict with a recent change that I made, it's for the better, but it'll probably be one of those merges where this change has to be thrown out and "re-applied" against a different baseline. It should be easy to do and I'm happy to help.

src/poprox_storage/repositories/accounts.py
kluver: I think we mostly export to _folders_ in existing buckets. I don't think you have access to see our current folder/bucket path so I recommend treating those like constant data / environment parameters so we can set the base paths as a final check after "does the code do what we need"

src/poprox_storage/repositories/panel_management.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class S3PanelManagementRepository(S3Repository):
+    def store_as_parquet(
+        self,
+        records: list[dict],
+        file_prefix: str,
+        start_time: datetime = None,
+    ):
+        return self._write_records_as_parquet(records, self.bucket_name, file_prefix, start_time)
```
kluver, line 13: I'm a little confused about what this function's purpose is -- I'm imagining it might make more sense in context which I just haven't gotten to in review yet. As it stands though -- it feels like this doesn't really do anything. It looks like platform will be responsible for converting accounts to dictionaries of metric values?

kluver (commented):
looking good, but the merge conflict is probably going to take cleaning. I left one comment.

What's needed to get this merge-ready from your perspectiev @sushobhan2024 ?

src/poprox_storage/repositories/panel_management.py
kluver: I don't think _anyone_ has this answer -- I think answering that question is part of the work of writing code for this PR. 

It just flagged my intuition to see a one-liner here, it makes me wonder if this function is needed, or if there's behavior that architecturally could be in here, but isn't. Where do we process model objects to output views in other exports? I could see that being a responsibility of this function. I authentically don't remember how we _normally_ split responsibilities for this work.

src/poprox_storage/repositories/accounts.py
kluver: I don't think we have an answer to this question as asked. 

Did you see the change that got merged earlier (the one that's marked as a merge conflict with this PR) I didn't _massively_ change the field list, the big thing I did is changed some architecture to move towards the same code patterns seen in other repos. This change makes a single function in charge of building most `Account` objects in practice. That change is for the better, but it'll conflict here since it moves the code around quite a bit.

Realistically, I _think_ we can then just do the code version of a select * from the table and let the code that builds the `Account` objects grab every row it wants/needs. That way we get a consistent property set.

karlhigley (approved):

PR #104, opened 2025-05-17
Set up storage for top news headlines

karlhigley:
Depends on https://github.com/CCRI-POPROX/poprox-concepts/pull/52

### DIFF
--- a/poprox-db/migrations/versions/2025_05_17_0754-1d20bc70fa6d_create_top_stories_table.py
+++ b/poprox-db/migrations/versions/2025_05_17_0754-1d20bc70fa6d_create_top_stories_table.py
@@ -1,0 +1,65 @@
+"""Create top_stories table
+
+Revision ID: 1d20bc70fa6d
+Revises: 45d10f860ff6
+Create Date: 2025-05-17 07:54:17.565202
+
+"""
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = '1d20bc70fa6d'
+down_revision: Union[str, None] = '45d10f860ff6'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "top_stories",
+        sa.Column("article_id", sa.UUID, nullable=False),
+        sa.Column("entity_id", sa.UUID, nullable=False),
+        sa.Column("headline", sa.String, nullable=False),
+        sa.Column("position", sa.Integer, nullable=False),
+        sa.Column("as_of", sa.DateTime, nullable=False),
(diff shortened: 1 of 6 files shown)


poprox-db/migrations/versions/2025_05_17_0754-1d20bc70fa6d_create_top_stories_table.py
```diff
@@ -0,0 +1,65 @@
...
+down_revision: Union[str, None] = '45d10f860ff6'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "top_stories",
+        sa.Column("article_id", sa.UUID, nullable=False),
+        sa.Column("entity_id", sa.UUID, nullable=True),
```
kluver, line 24: is the `entity_id` the topic for which this was listed? I felt like I understood this over in the concept, but I'm questioning myself now.

src/poprox_storage/repositories/articles.py
```diff
@@ -25,6 +26,8 @@
 
 NEWS_FILE_KEY = "mockObjects/ap_scraped_data.json"
 
+HeadlinePackage: TypeAlias = dict[str, dict[str, TopNewsHeadline]]
```
kluver, line 29: TIL. cool language feature

src/poprox_storage/repositories/articles.py
```diff
@@ -201,6 +206,24 @@ def fetch_article_by_url(self, article_url: str, newsletter_id: UUID | None = No
         else:
             return None
 
+    def fetch_entity_by_name(self, entity_name: str) -> Entity | None:
+        entity_table = self.tables["entities"]
+
+        entity_query = select(entity_table).where(entity_table.c.name == entity_name)
```
kluver, line 212: Do we currently expect to have/need entity name string normalization?

My gut says "no" -- but I've seen this sort of code be a constant pain point in tag-data (which definitely needs normalization) so I wanted to get a second check.

kluver (approved):
I didn't see anything blocking, but I left a few questions I wanted to double-check before merging.

poprox-db/migrations/versions/2025_05_17_0754-1d20bc70fa6d_create_top_stories_table.py
karlhigley: Yeah, we create an entity for every topic, so I figured it might be better to link them to the topic than repeat the topic name here

src/poprox_storage/repositories/articles.py
karlhigley: I think we're mostly okay because we're using data that's essentially exported from AP's knowledge graph. (Not saying it won't be an issue, but I'm comfortable ignoring it for now and waiting for it to bite us before trying to address it.)

PR #107, opened 2025-05-27
added fetch_feedback_impressions_by_account_id

Sris-Ty:


src/poprox_storage/repositories/newsletters.py
```diff
@@ -181,6 +181,61 @@ def fetch_impressions_by_newsletter_ids(self, newsletter_ids: list[UUID]) -> lis
...
+                impressions_table.c.impression_id,
+                impressions_table.c.newsletter_id,
+                impressions_table.c.article_id,
+                impressions_table.c.position,
+                impressions_table.c.extra,
+                impressions_table.c.headline,
+                impressions_table.c.subhead,
+                impressions_table.c.feedback,
+                articles_table.c.url,
+                articles_table.c.preview_image_id,
```
kluver, line 200: impressions also have a preview image id --  _think_ we should be using that one. 

What I don't know is if this column was "back-filled" or if the ideal version is to pull both and use the article preview_image_id as a fallback for the impression table preview_image_id when not set.

poprox-db/migrations/env.py
```diff
@@ -27,9 +27,9 @@
 # ... etc.
 
 db_user = os.environ.get("POPROX_DB_USER", "postgres")
-db_password = os.environ.get("POPROX_DB_PASSWORD", "")
+db_password = os.environ.get("POPROX_DB_PASSWORD", "thisisapoproxpassword")
 db_host = os.environ.get("POPROX_DB_HOST", "127.0.0.1")
-db_port = os.environ.get("POPROX_DB_PORT", 5432)
+db_port = os.environ.get("POPROX_DB_PORT", 5433)
```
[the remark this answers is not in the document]
Sris-Ty: Sorry i just mistakenly pushed this change too 😅.

src/poprox_storage/repositories/newsletters.py
kluver: So I think the design question is: do we backfill at the database, or do we assume null in the impression table means "load it from the articles table"

If we _plan_ to backfill at the database, can we code for that here and then spin up a second task to do that backfill eventually so we don't have to fix this later?

If we plan to load with default we can come back to that later I think, but it might be worth fixing now so we don't deploy a version of this code knowing it'll need fixing.

karlhigley (approved):

PR #109, opened 2025-06-08
Provide a foundation in the database for features that send multiple newsletters

karlhigley:
This makes schema changes we discussed on Friday to support multiple stories that involve sending additional newsletters to researchers and liaisons before and during experiments.

The schema we're aiming for looks like this, which generalizes the `recommenders` table, adds an `experiences` table, and connects newsletters to the experiences that generated them:

gnome-shell-screenshot-dt5gd4


kluver (approved):
Great start

PR #113, opened 2025-06-13
Set up verification token storage table and repository

karlhigley:
Depends on https://github.com/CCRI-POPROX/poprox-concepts/pull/57


kluver (approved):
I think this is a great start. I think this work misses/skips a pair of interesting questions:
 * Who actually _makes_ tokens (I.E. where is code-generation?)
 * Do we enforce expiration here or in the caller?
 * Do we want to make a table-cleanup function now, or wait until later?
 
 My instinct is to have code expiration and token generation in this file, I'm happy to look at those problems if you don't have ideas about how you want to handle them, and I think there's no reason those can't be a follow-up PR.

PR #115, opened 2025-06-16
put template into treatment table and manifest files.

kluver:


tests/repositories/test_experiments.py
```diff
@@ -91,3 +91,86 @@ def test_store_and_load_experiment():
 
         # with those adjustments -- assert the stored data matches the loaded
         assert loaded_experiment == experiment
+
+
+def _store_experiment(conn, path):
```
kluver, line 96: This cannot be directly used in the existing `test_store_and_load` function -- as it doesn't track some variables and do some assertions. If we _really_ wanted to it could be made to work but I think it's better to keep the two processes separate.

I'm also open to the comment "this is too lazy of a way to prep data for testing" but it was easy and it seems to work OK.

poprox-db/migrations/versions/2025_06_07_0854-87aa4d7933b2_create_an_experiences_table.py
```diff
@@ -46,7 +47,7 @@ def upgrade() -> None:
 
 
 def downgrade() -> None:
-    op.drop_constraint("fk_top_stories_team_id", "top_stories", type_="foreignkey")
-    op.drop_constraint("fk_experiences_recommender_id", "top_stories", type_="foreignkey")
+    op.drop_constraint("fk_experiences_team_id", "experiences", type_="foreignkey")
```
karlhigley, line 50: I think this also needs to change on line 41 above

karlhigley (approved):

poprox-db/migrations/versions/2025_06_07_0854-87aa4d7933b2_create_an_experiences_table.py
kluver: You're probably right. I'm surprised the tests din't catch that one. I wonder if our database schema tests are no longer automatic.

PR #116, opened 2025-06-22
fetch experience by id

Mahamudul42:
Added a function fetch_experience_by_id() for the poprox-platform call.

### DIFF
--- a/src/poprox_storage/repositories/experiments.py
+++ b/src/poprox_storage/repositories/experiments.py
@@ -4,6 +4,7 @@ from uuid import UUID
 from sqlalchemy import Connection, Table, and_, select, update
 
 from poprox_concepts import Account
+from poprox_concepts.domain import Experience
 from poprox_storage.concepts.experiment import (
     Assignment,
     Experiment,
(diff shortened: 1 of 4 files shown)


src/poprox_storage/repositories/experiments.py
```diff
@@ -152,6 +154,31 @@ def fetch_experiment_phases(self, experiment_id: UUID) -> list[Phase]:
 
         return list(phases.values())
 
+    def fetch_experience_by_id(self, experience_id: str) -> Experience | None:
+        # Fetch an Experience by its ID from the experiences table.
+        experiences_table = self.tables.get("experiences")
+        if not experiences_table:
```
karlhigley, line 160: You can assume the table will exist by this point in the code. If we failed to load it in the `__init__()` above, we will have already had an error by the time we get here

karlhigley (changes requested):
This code looks pretty good, but I don't think it belongs in this class. Instead I'd create a new `DbExperiencesRepository` class as a home for `fetch_experience_by_id()`

Mahamudul42:
Okay, I'll take care of it. I'll also update the poprox-platform codebase accordingly.

Mahamudul42:
Added an experience.py file for the DbExperiencesRepository class with the function fetch_experience_by_id.

Added a fetch_internal_accounts_by_team function in accounts.py

karlhigley (approved):

PR #117, opened 2025-06-23
Make experiment start and end dates inclusive

karlhigley:
This makes it easier to understand when the last newsletter for an experiment is going to go out: it goes out on the end date of the experiment, not the day before.


kluver (approved):
Logic is sound.

I'll be honest, I think I'll find inclusive-at-both-ends _more_ confusing since I feel more things are exclusive at the top-end in programming in general. At that point it's just a matter of preference though and I can work with either principal. If it does get confusing I bet we can do some minor renames and make it super unambiguous.

karlhigley:
What ended up being confusing with our first experiment is that the `end_date` of the concluding experiment and the `start_date` of a potential next experiment would have been the same.

PR #119, opened 2025-06-23
Return empty strings for recommend info field if not recorded

karlhigley:
This part of the record creation for the export fails with an error if `newsletter.recommender_info` is `None`. It _shouldn't_ be `None`, but if it is we should fail gracefully instead of preventing the export from running at all.


sushobhan2024 (approved):

sushobhan2024:
@karlhigley Thank you for the fix! What shall we do about the failing check?

PR #120, opened 2025-06-23
Add compensation method to the compensated activity export

karlhigley:
Didn't see this obviously missing field until we actually wanted to use the exported dataset 😅


kluver:
I don't know why it's always poprox-storage, but it's _always_ poprox_storage when one of us has a per-commit issue, I swear.

kluver (approved):

PR #123, opened 2025-07-01
Optimize the feedback fetching query

karlhigley:
This does a few different things, each of which reduces the estimated cost of the query a bit:
* Fetches all fields from the impressions table instead of just some
* Joins with the newsletters table first before joining with articles
* Moves impression sorting out of the query and into the surrounding code

### DIFF
--- a/src/poprox_storage/repositories/newsletters.py
+++ b/src/poprox_storage/repositories/newsletters.py
@@ -188,24 +188,16 @@ class DbNewsletterRepository(DatabaseRepository):
 
         query = (
             select(
-                impressions_table.c.impression_id,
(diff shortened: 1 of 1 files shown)


kluver (approved):
I'm no good at query optimizations so I'm going to have to trust you that the all-columns and join order things make a lick of difference, but the logic is sound for being equivalent.

Are prepared statements a thing? I seem to remember a lot of hem and haw about them when I was first getting my feet under me as a coder, but I haven't seen them in modern frameworks.

karlhigley:
Join order mostly doesn't matter (at least according to `EXPLAIN ANALYZE`.) The other parts seem to help a little bit though.

PR #125, opened 2025-07-01
Add a separate `top_story_packages` table

karlhigley:
This will allow us to keep track of which top stories came in the same bundle, which will simplify downstream queries because we can fetch the latest bundle for a particular entity (like `sports`.)

### DIFF
--- a/poprox-db/migrations/versions/2025_07_01_0957-d1f9922c638b_create_top_story_packages_table.py
+++ b/poprox-db/migrations/versions/2025_07_01_0957-d1f9922c638b_create_top_story_packages_table.py
@@ -1,0 +1,94 @@
+"""Create top story packages table
+
+Revision ID: d1f9922c638b
+Revises: 72901f4ab61c
+Create Date: 2025-07-01 09:57:31.375838
+
+"""
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = 'd1f9922c638b'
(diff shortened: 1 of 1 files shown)


poprox-db/migrations/versions/2025_07_01_0957-d1f9922c638b_create_top_story_packages_table.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def upgrade() -> None:
+    # Create the new table and link it to the entities table
+    op.create_table(
+        "top_story_packages",
+        sa.Column("package_id", sa.UUID, primary_key=True, nullable=False),
+        sa.Column("entity_id", sa.UUID, nullable=True),
+        sa.Column("as_of", sa.DateTime, nullable=False),
+        sa.Column("created_at", sa.DateTime, nullable=False, server_default=sa.text("NOW()")),
```
kluver, line 27: any particular reason to have `as_of` and `created_at` -- Like I think I can see why they would be different values in general, but having both feels a bit redundant.

kluver (approved):
lgtm

poprox-db/migrations/versions/2025_07_01_0957-d1f9922c638b_create_top_story_packages_table.py
kluver: Makes sense to me. :shipit:

karlhigley:
Same code, different file in order to avoid creating multiple heads in our Alembic revision graph

PR #135, opened 2025-07-14
Allow accounts to be updated at the model level.

kluver:
forked off of CCRI-POPROX/poprox-storage#134 but not doing the same thing. Only change local to this is the store_account change. 

change store_account function to be an upsert instead of direct insert, this suitably matches all current uses, and de-duplicates this from the new-account function below it.

I only made fields we might _want_ to update update, rather than all fields at once. I forwarded the `commit` property from the upsert function to allow web to say "no" and get it's exceptions for admin feedback purposes.

### DIFF
--- a/src/poprox_storage/repositories/accounts.py
+++ b/src/poprox_storage/repositories/accounts.py
@@ -72,20 +72,22 @@ class DbAccountRepository(DatabaseRepository):
             return accounts[0]
         return None
 
-    def store_account(self, account: Account) -> UUID | None:
+    def store_account(self, account: Account, commit=True) -> UUID | None:
         account_tbl = self.tables["accounts"]
-        query = (
-            sqlalchemy.insert(account_tbl)
-            .values(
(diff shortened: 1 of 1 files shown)


src/poprox_storage/repositories/accounts.py
```diff
@@ -56,44 +56,38 @@ def fetch_accounts_between(self, start_date, end_date) -> list[Account]:
...
+            {
+                "account_id": account.account_id,
+                "email": account.email,
+                "source": account.source,
+                "subsource": account.subsource,
+                "status": account.status,
+            },
+            # NOTE -- this is not explicitly named in out migrations and therefore is _fragile_
+            constraint="account_pkey",
+            commit=commit,  # do not commit so we throw errors.
```
karlhigley, line 89: This didn't commit before so it seems like this changes the behavior. Maybe the default should be `commit=False`?

karlhigley (approved):

PR #144, opened 2025-08-04
Raise error if a non-internal account is added to a team

kluver:
Considerations:

- Do we want a more specific error type here? If so, what? Do we have a PoproxError already? (I couldn't find one, but I could've missed it)
 - From what I can tell this is used in three clear places:
     1. Manual Ingress lambda for an experiment
     2. Admin Team membership interface
     3. Experimenter Team management interface
 - I would guess that 2. and 3. should be checked for decent error messages, but what about 1. Are we happy just letting this "fail fast?"

### DIFF
--- a/src/poprox_storage/repositories/teams.py
+++ b/src/poprox_storage/repositories/teams.py
@@ -2,7 +2,7 @@ from uuid import UUID
 
 from sqlalchemy import Connection, Table, select
 
-from poprox_concepts.domain.account import is_internal_account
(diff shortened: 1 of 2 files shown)


karlhigley (approved):
I think it's fine to let the experiment import Lambda fail fast

kluver:
Let's merge and do the web-side usability of this as a second pass. This won't "break" them, they'll just work in a hard-to-understand way.

PR #151, opened 2025-09-25
Entities code to support poprox entity page

Mahamudul42:


src/poprox_storage/repositories/account_interest_log.py
```diff
@@ -85,6 +86,103 @@ def fetch_topic_preferences(self, account_id: UUID) -> list[AccountInterest]:
         ]
         return results
 
+    def fetch_entity_preferences(self, account_id: UUID) -> list[dict]:
+        current_interest_tbl = self.tables["account_current_interest_view"]
+        entity_tbl = self.tables["entities"]
+
+        # Exclude topic entities that have separate preference forms
```
karlhigley, line 93: We should probably be selecting or excluding by type instead of a hard-coded list

kluver:
I guess I'm still unsure why we are excluding `topic`s here. Is there a reason for that _other_ than -- we don't want topics for the web? Otherwise it's rubbing me the wrong way. 

The cheapest solution would be to at least clearly document that the functions don't fetch topics. I'm imagining another developer working and seeing `fetch_entities_by_partial_name` and getting really confused why some entities are not being returned.

Mahamudul42:
I have an exam at the end of this week, so I created this PR early based on our discussion since we were aligned on the direction.

**Note: CI is currently failing due to what appears to be a poprox-concepts dependency caching issue, not related to the changes in this PR.**

### CI Issue Details:
- The CI environment installed an outdated version of poprox-concepts that doesn't include `TopNewsHeadline`
- `TopNewsHeadline` has been in poprox-concepts for a couple of months
- The import in `articles.py` existed before this PR - this is not new code

### Changes in This PR:
- Added `exclude_types` parameter to `fetch_entities_by_partial_name()` and  `fetch_entity_preferences()`
- Removed the `entity_type` column from `account_interest_log` (using JOIN with entities table instead)
- Makes the storage layer generic while allowing callers to control entity type filtering

I'll be available to address any feedback after my exam. 
@karlhigley @kluver

kluver:
Last changes look like they address my concern so I'm ready to see this batch of PRs merged.

PR #152, opened 2025-09-25
Mahamudul/features/image personalization experiment export

Mahamudul42:


karlhigley:
The part that's about images looks okay, but this also has code about other things in it

Mahamudul42:
Thanks for catching that. I’ve removed the entities code from here—it was added unintentionally due to the PR. I already addressed this in another PR, but it looks like it was left behind here as well.

src/poprox_storage/repositories/newsletters.py
```diff
@@ -315,6 +315,9 @@ def extract_and_flatten(newsletters: list[Newsletter], include_treatment: bool =
             record["profile_id"] = str(newsletter.account_id)
             record["newsletter_id"] = str(newsletter.newsletter_id)
             record["article_id"] = str(impression.article.article_id)
+            record["article_preview_image_id"] = (
+                str(impression.article.preview_image_id) if impression.article.preview_image_id else None
```
kluver:
```suggestion
                str(impression.preview_image_id) if impression.preview_image_id else ""
```
The rest of the code around this seems to use `""` as null instead of a true `None` -- I vaguely remember @karlhigley running into issues with literal `None` in reviews.

Suggested change also accounts for a change we made recently (adding a `preview_image_id` field to `impression`)

kluver (approved):
With one little change to fit recent modifications and handle null the same way as surrounding code (see suggested change) I think we're in business. Sorry I dropped the ball for basically the entire month of October.

src/poprox_storage/repositories/newsletters.py
karlhigley: The types in each column need to be consistent in order to successfully write Parquet files 👍🏻

Mahamudul42: @kluver 
I’ve updated the code based on the feedback. As of now, I haven’t identified any remaining issues.

PR #177, opened 2025-12-17
Rework `DbCandidatePoolRepo` to use `CandidateSet/Pool` concept model

karlhigley:
Based on https://github.com/CCRI-POPROX/poprox-concepts/pull/86

This is part of converging how the recommender and the platform think about candidate sets/pools, and a step toward including explicit candidate pools in our data exports.

### DIFF
--- a/src/poprox_storage/repositories/pools.py
+++ b/src/poprox_storage/repositories/pools.py
@@ -1,16 +1,17 @@
 import logging
-from datetime import date
-from uuid import UUID, uuid4
+from datetime import date, datetime, time
+from uuid import UUID
 
 from sqlalchemy import (
     Connection,
     Table,
+    and_,
     desc,
     select,
 )
 from sqlalchemy.dialects.postgresql import insert
 
-from poprox_concepts.domain import Article
+from poprox_concepts.domain import CandidatePool
 from poprox_storage.repositories.articles import _fetch_articles
 from poprox_storage.repositories.data_stores.db import DatabaseRepository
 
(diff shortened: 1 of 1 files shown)


src/poprox_storage/repositories/pools.py
```diff
@@ -25,42 +26,87 @@ def __init__(self, connection: Connection):
...
         candidate_pools_table = self.tables["candidate_pools"]
         candidate_articles_table = self.tables["candidate_articles"]
 
-        candidate_pool_id: UUID = uuid4()
-
-        set_insert_stmt = insert(candidate_pools_table).values(
-            {"candidate_pool_id": candidate_pool_id, "pool_type": pool_type}
+        pool_insert_stmt = (
+            insert(candidate_pools_table)
+            .values(candidate_pool_id=pool.pool_id, pool_type=pool.pool_type)
```
kluver: I can't remember -- don't we get a crash if we explicitly insert `candidate_pool_id=None`? I thought that's why we do in other places:
```
candidate_pool_id: UUID = pool.pool_id
if candidate_pool_id is None:
    candidate_pool_id = uuid4()
```

src/poprox_storage/repositories/pools.py
```diff
@@ -25,42 +26,87 @@ def __init__(self, connection: Connection):
...
+    def fetch_candidate_pool(self, candidate_pool_id: UUID) -> CandidatePool:
+        pools_table = self.tables["candidate_pools"]
+        candidates_table = self.tables["candidate_articles"]
         articles_table = self.tables["articles"]
         links_table = self.tables["article_links"]
 
+        # First get the pool attributes
+        pool_query = select(pools_table).where(pools_table.c.pool_id == candidate_pool_id)
+        row = self.conn.execute(pool_query).one_or_none()
+        pool = CandidatePool(pool_id=row.candidate_pool_id, pool_type=row.pool_type, created_at=row.created_at)
```
kluver: is it worth updating the type signature of this function to return `CandidatePool | None` and returning `None` here if not found instead of crashing?

I'm a bit torn -- failing fast here is useful.

src/poprox_storage/repositories/pools.py
```diff
@@ -25,42 +26,87 @@ def __init__(self, connection: Connection):
...
+
+        return pool
+
+    def fetch_candidate_pools_between(self, start_date: date, end_date: date) -> list[CandidatePool]:
+        pools_table = self.tables["candidate_pools"]
+        candidates_table = self.tables["candidate_articles"]
+        articles_table = self.tables["articles"]
+        links_table = self.tables["article_links"]
+
+        start_dt = datetime.combine(start_date, time.min)
```
kluver: cool

kluver (changes requested):
Logic makes sense to me, but I want a double-check on null-behavior in a few places, and there are a few things I think are minor typos.

kluver (approved):
lgtm.

Not going to press merge just now. Waiting until monday after the newsletters go out. That way we can test a rollup of this and any other PRs with enough breathing room to debug if needed.

PR #193, opened 2026-05-04
allow NewsletterRepo.fetch_newsletters_between to take a list of accounts or account_ids equally.

kluver:


AndrewNordstrom (approved):
lgtm, backward compatible and CI's green so approving.

one heads up: a non-empty list that's neither Accounts nor UUIDs (a stringified id, say) now silently skips the account filter and returns everything in the window, where it used to throw. nothing hits that path today, but might be worth raising on an unexpected type instead of failing open.

PR #197, opened 2026-06-24
Normalize organization entity_type spelling (organisation -> organization)

Mahamudul42:


kluver:
Ahh ... there it is. We should make sure this and CCRI-POPROX/poprox-platform#796 are noted as linked so that we can sequence them correctly

PR #198, opened 2026-07-07
make fetch_article_by_external_id source-aware

AndrewNordstrom:
Groundwork before ProPublica ingest (PROJ-1327). fetch_article_by_external_id only filtered on external_id, so once a second source writes rows with overlapping ids it could hand back the wrong source's article. this makes it source-aware via an optional arg (old callers unchanged) and adds a non-unique index on (source, external_id) for that lookup.

I didn't touch uq_articles or the write path: urls are source-specific (propublica.org vs apnews.com) so store_article already can't collide across sources, and (source, external_id) can legitimately repeat (kept as history, deduped to latest at read time) so a unique constraint would be wrong.

rebased onto current main and tested locally against a fresh dev db:
- [x] two sources sharing an external_id stay distinguishable; wrong-source / unknown-id -> None
- [x] omitting source keeps the old behavior (backward compatible)
- [x] metadata-only link-out article (body=None, no images/mentions) round-trips
- [x] same (source, external_id) keeps history, fetch returns the newest
- [x] full existing suite green; ruff + pre-commit clean; migration upgrade succeeds

one unrelated heads-up: the alembic stairstep test (poprox-db/test.sh, not in CI) has a pre-existing failure -- the create_article_packages_tables downgrade references "particle_ackage_contents" instead of "article_package_contents". not from this change, happy to fix separately if useful.

### DIFF
--- a/poprox-db/migrations/versions/2026_07_06_1830-62b01cd7d748_add_source_external_id_index_to_articles.py
+++ b/poprox-db/migrations/versions/2026_07_06_1830-62b01cd7d748_add_source_external_id_index_to_articles.py
@@ -1,0 +1,30 @@
+"""add source external_id index to articles
+
+Adds a non-unique index on articles(source, external_id) to back the source-aware
+lookup added to DbArticleRepository.fetch_article_by_external_id. The pair is
+intentionally NOT unique: multiple rows per (source, external_id) are expected --
+they are kept as history and deduped to the latest row at read time.
+
+Revision ID: 62b01cd7d748
+Revises: d4e5f6a7b8c9
+Create Date: 2026-07-06 18:30:32.656756
+
+"""
+
+from typing import Sequence, Union
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "62b01cd7d748"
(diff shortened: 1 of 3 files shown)


kluver:
>  happy to fix separately if useful.

Yeah... we've had a few of these slip by. I'm not entirely sure how -- I wouldn't be surprised if there was just one that went through without tests or something and then the rest are cascades from fixing that or something. SOP has been "jut fix them where you see them" especially since they have yet to be a real problem.

kluver:
> tested locally against a fresh dev db:

If you haven't gotten them working yet we can toy with it in a pair programming session. But there is a unit testing suite for poprox_storage, and some tests there wouldn't be bad. I don't think we've made it a requirement though.

kluver:
> > tested locally against a fresh dev db:
> 
> If you haven't gotten them working yet we can toy with it in a pair programming session. But there is a unit testing suite for poprox_storage, and some tests there wouldn't be bad. I don't think we've made it a requirement though.

And I see you've already written those now that I get into the actual line review

kluver (approved):
lgtml :shipit:

Solid tests. good comments and reasoning. overall happy. Merge when you see fit.

PR #199, opened 2026-08-19
Support article link data export

sushobhan2024:
Added `DbArticleRepository.fetch_all_article_links()`, and `S3ArticleRepository.store_article_links_as_parquet()` (platform code calls these).

### DIFF
--- a/src/poprox_storage/repositories/articles.py
+++ b/src/poprox_storage/repositories/articles.py
@@ -640,18 +640,6 @@ class S3ArticleRepository(S3Repository):
         records = extract_and_flatten_links(links)
         return self._write_records_as_parquet(records, bucket_name, file_prefix, start_time)
 
-    def extract_and_flatten_links(links: list[ArticleLink]) -> list[dict]:
-        """Flatten article_links rows for parquet export"""
-        return [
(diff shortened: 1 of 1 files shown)


src/poprox_storage/repositories/articles.py
```diff
@@ -374,6 +374,20 @@ def fetch_latest_package_by_entity(self, entity_id: UUID) -> ArticlePackage | No
             created_at=package_row.created_at,
         )
 
+    def fetch_all_article_links(self) -> list[ArticleLink]:
```
kluver: I'm concerned about this function signature -- there are _very_ few situations in which it feels right to fetch all links, instead of all links _for a given set of articles_

Certainly there will be a point at which this is simply too many article links to read, or we'll be fetching a lot of links we don't need right?

Maybe we could add a list of article ids (source) to fetch all links, or we could introduce variants of the function:
`fetch_all_article_links_from(article_ids)` and `fetch_all_article_links_to(article_ids)`?

kluver (changes requested):

kluver (approved):
lgtm

